PR #35, opened 2015-01-13
Added support for polygons in Ply files

dedalusj:
Add support for drawing polygons from a Ply file.

The format for defining polygons in the Ply file is a `polygon` element with the properties: 
- `outer_vertex_index` is a list of indices of vertices defining the outer ring of the polygon
- `inner_polygon_vertex_counts` is a list containing the number of vertices in each hole of the polygon
- `inner_vertex_index` is a list of all the indices of the vertices composing the inner polygons defining the holes

The implementation looks a bit "hacky" really but I couldn't find a nice way to go around the callback architecture.

The problem is that you need to know how many vertices each inner polygon/hole before you can populate the `edges` vector of the `PlyLoadInfo` struct. In order to do that however you need to parse the `inner_polygon_vertex_counts` property which doesn't happen until the end of `loadPlyFile`.

The solution was to define a generic callback for lists that load the necessary data in two temporary vectors and  reconstruct the inner polygons at the end after `ply_read` has been called.


c42f:
This looks pretty good.  I suspect there's a slightly cleaner way to lay out the parser code - I'll see if I can think of anything.

On a related note, I saw the following which is quite interesting: http://code.google.com/p/polypartition/ .  Polypartition looks quite lightweight, it could be worth taking as a bundled dependency (ie, in thirdparty/polypartition/*) and creating an actual triangulation rather than just an edge list.

src/mesh.cpp
```diff
@@ -235,6 +259,28 @@ bool loadPlyFile(const QString& fileName,
     }
     if (info.colors.size() != info.verts.size())
         info.colors.clear();
+
+    // Reconstruct inner polygons
+    while (innerPolygonVertexCount.size())
```
c42f: This would be cleaner as a for loop - no need to be modifying `innerPolygonVertexCount`, right?  (If you needed to do it this way, the conditional check would also be better written as `while(!innerPolygonVertexCount.empty())`)

src/mesh.cpp
```diff
@@ -235,6 +259,28 @@ bool loadPlyFile(const QString& fileName,
...
+    {
+        unsigned int verticesCount = innerPolygonVertexCount.back();
+        unsigned long firstIdx = info.edges.size();
+        for (size_t j = 0; j < verticesCount; ++j)
+        {
+            if (j > 1)
+                info.edges.push_back(info.edges.back());
+            info.edges.push_back(innerVertexIndices.back());
+            innerVertexIndices.pop_back();
```
c42f: Again, I think it would be more natural to iterate over `innerVertexIndices` (in forward order) rather than erasing elements from it.  Especially since the way it's written here the order of everything gets reversed by starting from the back.  It's not wrong, but I did find it confusing.

dedalusj: Agreed on using `while(!innerPolygonVertexCount.empty()))`.

I chose to do it in reverse because `innerPolygonVertexCount` and `innerVertexIndices` can potentially contain information about multiple holes or holes belonging to multiple polygons defined in the ply file.

Because of this I thought it was safer to pop entries off the "stacks" in the opposite order they were inserted. It seemed sligthly less error prone but I agree a little harder to understand.

I can add a comment about the logic or figure out the logic to do it forward.

c42f: This was what I had in mind:

```C++
    for (size_t i = 0, j = 0; i < innerPolygonVertexCount.size(); ++i)
    {
        size_t count = innerPolygonVertexCount[i];
        for (size_t k = 0; k < count; ++k)
        {
            info.edges.push_back(innerVertexIndices[j + k]);
            info.edges.push_back(innerVertexIndices[j + (k+1)%count]);
        }
        j += count;   
    }
```

dedalusj:
That is indeed a much cleaner way of doing it, great suggestion.

Sorry it took me a few days to make the changes I wanted to try it first and didn't have a Linux machine at home.

c42f:
No problems at all, I've been a bit slow reviewing this one.

It's becoming clear that the details of the ply format is turning into a kind of geometry API for displaz.  I'm afraid it's not a great API - I'm going to have to do some thinking about what to replace it with :-/

Anyway, I'm happy to have this patch, thanks a lot!

Just a note about branches - next time you submit a pull request, can you do it from an aptly named branch in your own repo rather than master?  That way it's clearer when merging from the command line.  More importantly, if you want to make two pull requests at the same time it avoids them getting mixed together.

PR #39, opened 2015-02-03
Set ilmbase variables when not BUILD_EXTERNAL

gadomski:
This patch includes a version-enabled find module for ilmbase.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,8 @@ if (DISPLAZ_BUILD_EXTERNAL)
         set(LASLIB_LIBRARIES ${THIRDPARTY_INSTALL_DIR}/lib/${CMAKE_STATIC_LIBRARY_PREFIX}laslib${CMAKE_STATIC_LIBRARY_SUFFIX})
     endif()
     add_subdirectory(thirdparty/external)
+else()
+    find_package(Ilmbase 2.0.0 REQUIRED)
 endif()
 
 if (DISPLAZ_USE_LAS)
--- a/cmake/FindIlmbase.cmake
+++ b/cmake/FindIlmbase.cmake
@@ -1,0 +1,29 @@
+# Find ilmbase
+#
+# This sets:
+#   - ILMBASE_FOUND:  system has ilmbase
+#   - ILMBASE_INCLUDES: the ilmbase include directories
+#       (note that this is a different name than is conventional for
+#        find modules, to stick with existing project norms)
+#   - ILMBASE_LIBRARIES: the ilmbase libraries
+#   - ILMBASE_VERSION: the version string for ilmbase
+
+find_path(ILMBASE_INCLUDE_DIR OpenEXR/IlmBaseConfig.h)
+find_library(ILMBASE_IMATH_LIBRARY Imath)
+find_library(ILMBASE_IEX_LIBRARY Iex)
+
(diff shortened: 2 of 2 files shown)


c42f:
Looks great, I like the `find_package_handle_standard_args` stuff.

I did some testing of older versions, since ubuntu (and presumably other debian based linux distros) still come with the ancient ilmbase-1.0.1.  Surprisingly that really old version seems to work fine so I downgraded the required version and fixed the cmake version detection for versions older than 2.0.0.

I honestly couldn't remember why I put ilmbase in the third party build system in the first place... c186fd says it was because I wanted the windows build to be less painful.

gadomski:
Right on. Yeah sometimes the vendor-vs-make-the-user-do-it decision can be tricky. How did you like using cmake's external project infrastructure? Would you recommend?

c42f:
I'd cautiously recommend it, depending on what you're trying to do.  For building obscure or new libraries without package manager support, or libraries without good build systems of their own it's pretty great.

Also if you really want to pin down your library dependencies and only upgrade in a well controlled way it's good.  I've used it to automate the build of a moderately large collection of libraries at work basically for this reason.  The following come out of the box:
* Support for specifying arbitrary dependencies
* Great parallel build support, given the third party builds are make or cmake-based.
* Nice support for downloading packages from a URL and verifying checksums
* Support for custom patch steps (though you need an external patch.exe on windows or some such thing)
* Support for arbitrary custom build steps for the odd super-funky workaround

All of this is basically built in, so there's very little logic to write even for a solid production build system of third-party libs with the various strange hacks that usually requires.

Where it falls down a bit is integrating it into the main project build system as I tried to do in displaz.  Since cmake doesn't really *understand* the third party build systems which it's trying to wrap up - it doesn't know about which targets they build - it can't scan dependencies properly, find libraries with the correct names, etc and you end up with manual specification of dependencies between the external project targets and the actual built targets of the main build.

You may have noticed some of the resulting mess, for example manually constructing library names since find_library won't work, ugh.  To fix that I'd need to get the third party libs built and installed before the main cmake runs.  The simplest option is to have the third party build system separate, and have people run it manually before they run the main build.  Another option is a meta build, where the main project is just one of several external projects to be built.

gadomski:
Yeah, the manual construction of library names is a bit janky. I wonder if you can use `find_package(MyDep REQUIRED PATHS "${THIRDPARTY_INSTALL_DIR}/mydep/lib" NO_DEFAULT_PATH)` to let it search for the package and set things up for you? Just guessin', haven't tried.

Thanks for the beta on the external project stuff. Seems like a specific use case for strange, niche packages that can't just be vendored (included in the source tree) straight-up?

c42f:
Problem is, the external package build systems run at build time, not at cmake time, so there's nothing there for `find_package` to find.

It's definitely an alternative to just include third party packages in the source tree if they're small enough.  Maybe I should just do that with the parts of lastools which I actually use (lastools packages aren't versioned in the URL, which makes things rather hard on the build automation side of things).

gadomski:
Ah yesh, I forgot about that detail.

Yeah, lastools is a funny one to depend upon.

c42f:
I'd use liblas instead - just trying to avoid the boost dependency.

gadomski:
:+1:


-- 
Pete GadomskiPhysical Scientist

Cold Regions Research & Engineering Lab

pete.gadomski@gmail.com

530.440.4630

PR #41, opened 2015-05-15
A matlab function to get cursor position from displaz

PaulBellette:
A simple system call and string parse deal to get the displaz cursor psoition

### DIFF
--- a/bindings/matlab/get_dcursor_pos.m
+++ b/bindings/matlab/get_dcursor_pos.m
@@ -1,0 +1,26 @@
+function pos = get_dcursor_pos()
+
+%Get the position of the point under the current displaz cursor focus
+
+fixLdLibPath = '';
+if isunix()
+    % Reset LD_LIBRARY_PATH on unix - matlab adds to it to get the
+    % matlab version of libraries, which tend to conflict with the
+    % libraries used when building displaz.
+    fixLdLibPath = 'env LD_LIBRARY_PATH=""';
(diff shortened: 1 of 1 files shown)


bindings/matlab/get_dcursor_pos.m
```diff
@@ -0,0 +1,24 @@
...
+end
+
+[exitStatus, commandStdout] = system(sprintf('%s displaz -querycursor', fixLdLibPath));
+
+if exitStatus ~= 0
+    errorStr = [];
+    errorStr = strcat(errorStr, sprintf('ERROR: Failed to get displaz cursor position.  Stdout was:\n'));
+    errorStr = strcat(errorStr, sprintf('\n%s\n',commandStdout));
+    errorStr = strcat(errorStr, sprintf('\nERROR: Command Line used was:\n'));
+    errorStr = strcat(errorStr, sprintf('\n%s\n',commandLine));
```
c42f: Thanks for the PR!  Note that commandLine isn't defined so the error will error.

bindings/matlab/get_dcursor_pos.m
```diff
@@ -0,0 +1,24 @@
...
+    error(errorStr);
+end
+
+pos = sscanf(commandStdout,'%f');
```
c42f: The rest of the displaz interface assumes xyz are rows, so better to transpose this.

PR #46, opened 2015-06-08
Fixup: Cursor center point flickers due to depth testing

nigels-com:
Seems to me the cursor center shouldn't flicker, depending on the result of depth testing.
This small patch disables depth testing for the 3D cursor center.

### DIFF
--- a/src/view3d.cpp
+++ b/src/view3d.cpp
@@ -421,6 +421,9 @@ void View3D::drawCursor(const TransformState& transStateIn, const V3d& cursorPos
     if((relCursor * transState.modelViewMatrix).z > 0)
         return;
 
+    glPushAttrib(GL_ENABLE_BIT);
+    glDisable(GL_DEPTH_TEST);
+
     if (centerPointRadius > 0)
     {
         transState.load();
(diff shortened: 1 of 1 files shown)


c42f:
The depth testing of the cursor center was meant to be a way of giving the user some depth information.  I guess that's not obvious though - maybe there's a better way to do it?

nigels-com:
So the idea is that the dot is shown unless there is other geometry occluding it from the viewer?

nigels-com:
The current concept of a selection seems to limited to a selected point in the dataset.  But the cursor is free to move in space, right?  If we draw the cursor with a less-or-equal depth test, and use the same sprite rendering mode, the fragments ought to match, eventhough the color/texture/stipple can be different.

c42f:
Yes, the dot is drawn at the 3D location of the cursor so that it's visible unless it's behind other geometry.  The cursor is actually free to move in space (ctrl+LMB+drag) so it can actually be off a vertex of the dataset, but I don't have a compelling use case for that, so it's usually set to the camera centre of rotation at the moment.

You may have seen that I'm writing to `gl_FragDepth` in the fragment shader for the default marker shape so that the points come out as little spheres when they're greater than a couple of pixels on the screen.  I'm afraid that makes the result of the point vs cursor depth test depend on the relative zoom.

nigels-com:
Oh, I see.
I'll retreat on this pull request, if you don't mind.

c42f:
Of course that's fine.  Though if you got confused by the intent of this bit of UI that's not a good sign that it's "right".  Suggestions on how to handle it better are very welcome.

PR #47, opened 2015-06-09
Implement 3D axis direction legend, UI integration to toggle on or off.

nigels-com:


c42f:
Thanks, this is cool.  I particularly like the semitransparent circle to add contrast.  I'd be inclined to tweak it a bit for contrast and antialiasing of the outside line - perhaps we can do that tomorrow.

One thing I did notice is that the axes aren't drawn in the same perspective as the rest of the 3D view.  I find this really visually confusing/disturbing since the vanishing point of the axes is inconsistent with the rest of the scene, for example the bounding boxes.  This is the main thing that I think really needs changing.

I'm not entirely sold on embedding the textures in C files the way you do, only because it's so verbose even for such tiny textures, and for anything much larger would really bloat the executable size.  There's a couple of almost-as-simple alternatives I can see if we save them in any image format Qt can open (png for example):
* Install all textures when installing displaz, using the cmake `install(FILES ...)` signature.  Add the textures directory to the `QFile` search path (see `setupQFileSearchPaths()`).
* Get Qt to bundle them into the executable, using the cmake macro `qt4_add_resources()`.  You should then be able to open them using `QFile` (see here: http://doc.qt.io/qt-5/resources.html).  [Side note: if the qt resource system didn't do it already, `xxd -i axes.png` is a nice way to generate embeddable binary data in C source form for arbitrary files.]

nigels-com:
Perspective projects means that parallel lines do not remain parallel.  (Ortho for the win!)  What's being shown is based purely on the camera view matrix.  I'm not sure that factoring the projection matrix will help, but I'll look into that.

Smoothing the edge of the circle can be done in GIMP, my feeling was that a blurred edge was a little less pleasing visually, but I'll look into that too.

I'll also look into QFile and xxd for the textures.  It would be nice if these were zlib compressed, at least.

It would be nice

c42f:
With QFile, and QImage, you can load pngs directly from embedded resources.  (The axes texture becomes 612 bytes if saved as a png!)

I get the stuff about the parallel lines, but there's no getting around the massive downside: ortho projections give no sense of depth.

I think the issue is the following: Any axes projected to a *particular x,y location in the 2D image plane* agree in their three principle directions, regardless of their depth into the scene.  So at a given x,y location (the bottom left corner in this case), there's a uniquely defined set of directions which yield the same vanishing points as the rest of the 3D geometry in the scene.  Using the view matrix only corresponds to choosing the centre of the screen as the location for the axis widget (but obviously this isn't where we are putting it).

nigels-com:
I'll leave the perspective axes problem as "to be decided".  The general direction of x, y, and z are faithfully indicated, at least!

nigels-com:
Freshly rebased and squashed into one commit.

Cheers,
Nigel

src/glutil.h
```diff
@@ -105,6 +107,46 @@ inline void glLoadMatrix(const Imath::M44f& m)
...
+//------------------------------------------------------------------------------
+// Texture utility
+
+struct Texture
+{
+    Texture(const QImage& i)
+    :   image(i),
+        target(GL_TEXTURE_2D),
+        texture(0)
+    {
```
c42f: Shouldn't this check `i.format() == QImage::Format_ARGB32` or some such, just as a precaution?  If it's not in the right format, `image = image.convertToFormat(...)` should fix that.

I had a read through the qt source code, and `QGLContext::bindTexture()` has a whole texture cache and everything, based off cache keys held by `QImage`.  It's a bit complex and magical though, so I think I prefer what you have here.

c42f:
Thanks, the resource handling is much neater now.  I'll get back to this as soon as I can and solve the perspective axis issue before merging, unless you're inclined to have another look at that.

src/glutil.h
nigels-com: Good catch!  Yes, it makes sense to check the format.

Seems kinda lame to need a QGLContext reference to bind a texture, I'd put that down to an OpenGL/Qt paradigm mismatch.  I'd rather skip that path, unless you feel strongly about it.

I pushed a fixup for checking the format.  On the conversion, I generally wouldn't expect a 1-channel texture to be bloated to RGBA by default, so I'd be a bit more inclined to have an assertion there.

c42f: Agreed about the Qt-OpenGL mismatch.  The code looks good now thanks.

c42f:
I implemented a change to make the axes perspective-correct, and I think this helped.

I also tweaked the style a bit to make them more unobtrusive.  Not sure I've really got it right, but I want them to be just visible enough so that you can tell the orientation without being distracted by them.

The additional changes are on my 3d-axis-widget branch if you feel like checking them out.  Barring extra discussion I'll probably merge them tomorrow.

nigels-com:
Seems to me making the style less "obtrusive" undermines the legibility for a grey or white background.

c42f:
Yes, you quite right - it's definitely hard to read over the top of a busy background.  I'm not convinced this is a major problem though: the axis widget is something I want rarely; most of the time I just want to see the data and not be distracted by extra visual clutter.  OTOH, on the occasion when I want extra information it will be very nice to have the axis in the corner.

How about this for an option: by default, the axis is very transparent and unobtrusive (could be even more transparent than now).  However, when the user hovers the mouse over the axes, the transparency is dialed right back off.  This optimizes for the usual case ("I just wanna view the data itself"), but lets you see the axis when it matters.

nigels-com:
Sure.  Providing we can assume a high frame rate, and can afford to redraw for "mouse focus" events on the 2D widget.  Is that assumption valid?

c42f:
Yes, an adequate framerate is a valid assumption - displaz adaptively decides on how much geometry to render each frame, aiming at a given frame rate (IIRC it's set to aim at 20fps at the moment).

If the user stops moving the mouse, the remaining geometry is filled in incrementally by rendering to an application controlled framebuffer (`m_incrementalFramebuffer`).  This is blitted to the backbuffer before rendering the overlay and light weight geometry on top at each frame.

c42f:
Ok, I've merged this - the code seems good and additional style tweaking can always be done later.  Cheers!

Issue #44, opened 2015-06-06
Color maps as user-editable shader parameters


c42f:
It would be cool to have color maps exposed as user-editable shader parameters in the same way as floats and ints are now.  Presumably these would just be sampler1D uniforms in the shader.  On the UI side, a combo box of standard colormaps might do, with an option to load custom ones from an image file.

Issue #43, opened 2015-06-05
User-editable edge and face shaders


c42f:
At the moment, the user can only tweak the shader for point clouds.  It would be great if the edge and face shaders were also available for editing.  This will require some UI work to connect the shader editor up to multiple loaded shaders.

Issue #15, opened 2013-08-22
First person camera and control scheme


c42f:
For the FPS players out there who undoubtedly won't stop requesting this if it's not supported... and for the potential Oculus Rift support which will require something similar.

nigels-com:
Do we know anything about where the ground is?  Which way is up?

c42f:
Yes, z is usually up and I think assuming this will be fine for now.  For the cases where it's not, I currently include a trackball camera option (see View->Use trackball camera).

landersson:
I'd love to see FPS style navigation added to displaz.

Issue #12, opened 2013-08-22
Trim view based on line segment


c42f:
The radial trimming is cool, but users often want to do it using a segment instead.  Ideally should be able to specify the segment with two clicks and a drag from an orthographic view.

PR #48, opened 2015-06-09
New routine for closest point search

gphysics:
Both old and new methods called for timing comparisons.  Need to delete unnecessary call to old method.

### DIFF
--- a/src/pointarray.cpp
+++ b/src/pointarray.cpp
@@ -61,69 +61,46 @@ struct OctreeNode
             delete children[i];
     }
 
-/*    void print(const int indent, const bool treeWalkDown, const V3f *const P) const
-    {
-        char s[256];
-        for(int i = 0; i < indent; i++)
-            s[i] = ' ';
-         s[indent] = 0;
-        printf("%sCenter = (%f %f %f)\n", s, center.x, center.y, center.z);
-        printf("%sradius = %f\n", s, radius);
-        printf("%sCenter = (%f %f %f)\n", s, bbox.center().x, bbox.center().y, bbox.center().z);
-        printf("%sBox    = (%f %f %f)\n", s, bbox.size().x, bbox.size().y, bbox.size().z);
-        printf("%sbeginIndex = %lu\n", s, beginIndex);
-        printf("%sendIndex   = %lu\n", s, endIndex);
-
-        if(endIndex > beginIndex)
-        {
-            V3f xMin( FLT_MAX,  FLT_MAX,  FLT_MAX);
-            V3f xMax(-FLT_MAX, -FLT_MAX, -FLT_MAX);
-            for(size_t i = beginIndex; i < endIndex; i++)
-            {
-                const V3f& p = P[i];
-                xMin.x = p.x < xMin.x ? p.x : xMin.x;
-                xMin.y = p.y < xMin.y ? p.y : xMin.y;
-                xMin.z = p.z < xMin.z ? p.z : xMin.z;
-                xMax.x = p.x > xMax.x ? p.x : xMax.x;
-                xMax.y = p.y > xMax.y ? p.y : xMax.y;
-                xMax.z = p.z > xMax.z ? p.z : xMax.z;
-            }
-            printf("%sxMin   = (%f %f %f)\n", s, xMin.x, xMin.y, xMin.z);
-            printf("%sxMax   = (%f %f %f)\n", s, xMax.x, xMax.y, xMax.z);
-        }
-
-        if(treeWalkDown)
-            for (int i = 0; i < 8; ++i)
-                if(children[i] != 0)
-                {
-                    printf("%sChild[%i]:\n", s, i);
-                    children[i]->print(indent + 2, treeWalkDown, P);
-                }
-    }*/
-
     bool rayPassesNearOrThrough(
-        const V3f& rayOrigin,
-        const V3f& rayDir) const
+        const V3d& rayOrigin,
+        const V3d& rayDir) const
     {
-        const float diagRadius = 1.2*bbox.size().length()/2; // Sphere diameter is length of box diagonal
+        const double diagRadius = 1.2*bbox.size().length()/2; // Sphere diameter is length of box diagonal
 
-        V3f o2c = bbox.center() - rayOrigin; // vector from rayOrigin to box center
+        V3d o2c = bbox.center() - rayOrigin; // vector from rayOrigin to box center
 
-        const float l = o2c.length();
+        const double l = o2c.length();
 
         if(l < diagRadius)
             return true; // rayOrigin lies within bounding sphere
 
         // rayOrigin lies outside of bounding sphere
 
-        const float cosA = (o2c ^ rayDir)/l;  // cosine of angle between rayDir and vector from origin to center
+        const double cosA = (o2c ^ rayDir)/l;  // cosine of angle between rayDir and vector from origin to center
 
-        if(cosA < FLT_MIN)
+        if(cosA < DBL_MIN)
             return false; // rayDir points to side or behind with respect to direction from origin to center of box
 
-        const float sinA = sqrtf(1 - cosA*cosA); // sine of angle between rayDir and vector from origin to center
+        const double sinA = sqrt(1 - cosA*cosA); // sine of angle between rayDir and vector from origin to center
 
         return sinA/cosA < diagRadius/l;
+    }
+
+    size_t closestPoint(
+        const V3f *const p,
+        const V3f& rayOrigin,
+        const V3f& rayDirection,
+        const double longitudinalScale,
+        double& thisRClosest) const
+    {
+         // calls utils closestPointToRay for the points within this octree node
+         return beginIndex + closestPointToRay(
+                                 p + beginIndex,
+                                 endIndex - beginIndex,
+                                 rayOrigin,
+                                 rayDirection,
+                                 longitudinalScale,
+                                 &thisRClosest);
     }
 
     size_t size() const { return endIndex - beginIndex; }
@@ -234,8 +211,7 @@ static OctreeNode* makeTree(int depth, size_t* inds,
 PointArray::PointArray()
     : m_npoints(0),
     m_positionFieldIdx(-1),
-    m_P(0),
-    m_lastVertexPickIdx(0)
+    m_P(0)
 { }
 
 
@@ -448,22 +424,21 @@ bool PointArray::loadFile(QString fileName, size_t maxPointCount)
 }
 
 
-V3d PointArray::pickVertex(const V3d& cameraPos,
-                           const V3d& rayOrigin, const V3d& rayDirection,
-                           double longitudinalScale, double* distance,
-                           std::string* info) const
+bool PointArray::pickVertex(const V3d& cameraPos,
+                            const V3d& rayOrigin,
+                            const V3d& rayDirection,
+                            const double longitudinalScale,
+                            V3d& pickedVertex,
+                            double* distance,
+                            std::string* info) const
 {
-    // NOTE rayOrigin here is not cameraPos but something else
-    // This routine uses cameraPos as the rayOrigin point
-
     if (m_npoints == 0)
-        return V3d(0);
+        return false;
 
-    const V3f cameraPosOffset = cameraPos - offset();
-    const V3f rayOriginOffset = rayOrigin - offset();
+    const V3d rayOriginOffset = rayOrigin - offset();
 
-    float rClosest  = FLT_MAX;
-    size_t iClosest   = 0;
+    double rClosest = DBL_MAX;
+    size_t idx = 0;
 
     std::vector<const OctreeNode*> nodeStack;
     nodeStack.push_back(m_rootNode.get());
@@ -472,7 +447,7 @@ V3d PointArray::pickVertex(const V3d& cameraPos,
         const OctreeNode* node = nodeStack.back();
         nodeStack.pop_back();
 
-        if(!node->rayPassesNearOrThrough(cameraPosOffset, rayDirection))
+        if(!node->rayPassesNearOrThrough(rayOriginOffset, rayDirection))
             continue;
 
         if (!node->isLeaf())
@@ -487,21 +462,18 @@ V3d PointArray::pickVertex(const V3d& cameraPos,
         }
 
         double thisRClosest;
-        size_t thisIdx = closestPointToRay(m_P + node->beginIndex, node->endIndex - node->beginIndex, rayOriginOffset, rayDirection,
-                         longitudinalScale, &thisRClosest);
+        size_t thisIdx = node->closestPoint(m_P, rayOriginOffset, rayDirection, longitudinalScale, thisRClosest);
         if(thisRClosest < rClosest)
         {
             rClosest = thisRClosest;
-            iClosest = thisIdx + node->beginIndex;
+            idx = thisIdx;
         }
     }
 
-    size_t idx;
+    if(rClosest == DBL_MAX)
+        return false;
 
-    if(rClosest == FLT_MAX)
-        idx = m_lastVertexPickIdx; // no nearest point to axis found either
-    else
-        idx = iClosest;
+    pickedVertex = V3d(m_P[idx]) + offset();
 
     if (info)
     {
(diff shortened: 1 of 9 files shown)


src/pointarray.cpp
```diff
@@ -44,8 +44,8 @@ struct OctreeNode
     size_t endIndex;         ///< End index of points in this node
     mutable size_t nextBeginIndex; ///< Next index for incremental rendering
     Imath::Box3f bbox;       ///< Actual bounding box of points in node
-    V3f center;              ///< Centre of the node
-    float radius;            ///< Distance to centre to node edge along an axis
+    V3f center;              ///< center of the node
```
c42f: Ok, I submit to the American spelling :)

(I do agree it's much better to be consistent here.  American spelling is so ingrained in APIs in general that I've started just using it in code.)

nigels-com: For a while I had a Colour class that was closely tied to an OpenGL glColor, which seemed increasingly discordant over time.  Then when things get data-driven via XML or JSON...

c42f:
Just to link things up - this looks like it'll address #11.

I'm going to look at this in detail later tonight or tomorrow - need to get some other work done for now.

Initial quick review: please clean it up by removing all debug code and making sure the old code path isn't executed.  (I'm happy to have some debug utilities like octree dumping, as long as they're cleanly implemented in separate functions and don't get called outside of debug mode.)

src/pointarray.cpp
c42f: Yes, it's slightly depressing, but crushed by the weight of 1000 APIs the only real option seems to be to give in :-)

gphysics:
Have cleaned up the code - no longer computes calculation time and only calls the old routine when the new one fails to find a point.

c42f:
By the way - I forgot to say, but this is great, thanks!  I've been meaning to get to it for ages but it's just been one of those things which got put off.

I did compile and check it out quickly in conjunction with the `selection` branch which I've been playing with - it's really needed in that case, since the selection code runs once per frame while painting selections on the point cloud.  One thing I noticed is that the lookups are usually lightning fast, but sometimes it does fall back to the old method and the classification painting becomes a bit unusable. 

Any idea why the new version sometimes fails to find a point?

nigels-com:
A vanilla octtree traversal would only be visiting nodes along the ray, not necessarily nodes in the neighborhood of the ray.  So depending on the octtree subdivision and relative size of voids, you could possibly not visit any leaf-nodes with any points?  (Just guessing)

nigels-com:
Right.  If we had an AAB (Axis Aligned Box) - Cone intersection test for rayPassesThrough we'd get more leaf nodes.  But there is a still a chance that the cone doesn't happen to touch any non-empty octtree leaf nodes.  In that case, try a bigger cone?  And then a bigger one again?

gphysics:
The routine tests each octree box to see if the ray passes though the sphere that perfectly contains the box (diameter = diagonalLength). So in most cases this will select boxes even when the ray just misses it. The main reason why the routine sometimes fails to find nodes is that for all points it tests the angle discrepancy between the vector (point - rayOrigin) and rayDirection to see if it is less than some threshold (maxTanA) - i.e. it finds points within a narrow cone around the rayDirection. If then returns the point within this cone that is closest to the rayOrigin. Presently this cone is set to be pretty narrow, and sometimes no points are found and the code then defaults to the previous algorithm. Widening the code will help, but then it may select points that are close but off-dir rather than points that are perfectly under the mouse. I think we will be able to improve this algorithm further - the fundamental speed up is there but the decision making process as to which point to could be improved. I'd like to put a bit more time in on this.

gphysics:
One more thing - it would be easy to omit the fallback to the previous look-up routine. The user can just keep moving the mouse until it does find a point!

gphysics:
I think I've nailed it now. The closestPoint() routine now finds (a) the closest point to the origin that is within a cylinder of radius 0.1m around the rayDir, and (b) the point closest to the rayDir vector. The routine will use the point from (a) if one exists, otherwise the point from (b). However it is still possible that it will not find a point (e.g. if the user clicks way off in the blue yonder). In this case it keeps the previous point. Chris - does the size of the squares change to suit the data spacing? If so I need to make the 0.1m a variable that is calculated.

src/pointarray.cpp
```diff
@@ -387,10 +506,53 @@ V3d PointArray::pickVertex(const V3d& cameraPos,
                            double longitudinalScale, double* distance,
                            std::string* info) const
 {
+    // NOTE rayOrigin here is not cameraPos but something else
+    // This routine uses cameraPos as the rayOrigin point
```
c42f: In the previous system, the `rayOrigin` is a rough guess at the depth of similar points the user is looking at in the scene.  In principle this is context-dependent, but on the current master branch it's basically a point on the clicked ray, at the depth of the centre of camera rotation.

The distinction between `cameraPos` and `rayOrigin` is confusing, I think I could clean things up here with a bit of refactoring.

c42f: (The idea was, if you can guess roughly where the user is looking, you should probably select something at a similar depth into the scene.)

c42f:
> Chris - does the size of the squares change to suit the data spacing? If so I need to make the 0.1m a variable that is calculated.

I'm not sure I quite understand this question about squares (cubes?), but here's some facts which may help answer it:
* Displaz currently deals fairly gracefully with high density data (for example, the space shuttle scan from here: plas.io/data/spaceshuttle.laz has one point per 5mm or so).
* The octree keeps subdividing until there's a given number of points in the leaves.  This is independent of data density.

c42f:
I've spent some time testing by clicking around.  The algorithm is really fast on large data which is great.  However, in the current form it lacks precision: when zooming in close, I find it quite difficult to select a particular point by clicking directly on it - the cursor often jumps to something nearby instead.  I think there's still some work to do here.  Happy to discuss/fix the issues in person tomorrow if you like.

src/pointarray.cpp
```diff
@@ -387,10 +506,53 @@ V3d PointArray::pickVertex(const V3d& cameraPos,
...
+        if (!node->isLeaf())
+        {
+            for (int i = 0; i < 8; ++i)
+            {
+                OctreeNode* n = node->children[i];
+                if (n)
+                    nodeStack.push_back(n);
+            }
+            continue;
+        }
```
c42f: This traversal seems to be out of the spirit of using a tree data structure: it's running `closestPoint` on *all* leaf nodes, rather than pruning early where possible based on the bounding boxes of the interior nodes.

Asymptotically this doesn't scale well.  It won't be noticeable with the really large leaf size I've got currently, but if the leaf size was made smaller in the future it might be a problem.  It seems a pity not to use the tree if it's available - I suspect you can just pull the first part of `closestPoint` out, and evaluate it for interior nodes.  If an interior node isn't in the path of the ray, you can discard it + all its children right away by not putting them onto the node traversal stack.

src/pointarray.cpp
```diff
@@ -387,10 +506,53 @@ V3d PointArray::pickVertex(const V3d& cameraPos,
...
+        node->closestPoint(m_P, cameraPosOffset, rayDirection, rToAxis, xToOrigin, iToAxis, iToOrigin);
+    }
+
+    size_t idx;
+
+    if(xToOrigin == FLT_MAX)
+    {
+        // no points within cylinder
+        if(rToAxis == FLT_MAX)
+            idx = m_lastVertexPickIdx; // no nearest point to axis found either
```
c42f: I don't think having `m_lastVertexPickIdx` here really makes sense - it's papering over the fact that nothing valid was found, and just returning something anyway.

After some thought, I'd say there's a couple of viable alternatives:
* Change the semantics of `pickVertex()` so that there's a way of signaling that no valid vertex was found.  The calling code then has to deal with it.  (Does it make sense to return a point when all vertices are behind the camera?  Kinda not, but it does allow the user to bring the data into view regardless of how they've messed up their camera.) 
* Change the code above so that the tree traversal happens in order of octree node proximity to the ray, and proceeds until it can be proven that the remaining nodes are further away than the currently selected closest point.  I favour doing this regardless, since it allows some neat things with selection painting which I can show tomorrow.  It should allow the previous selection distance algorithm to be accelerated in the octree framework.

gphysics:
In my last commit I merged the original rayPassesNearOrThrough test for the octree nodes into the closestPoint routine so that for each octree node it will only test the points if the ray passes near or through it. All octree nodes must be tested, but only some of them will test their points. If the user clicks in the blue yonder then no octree node will test its points, and no new vertex will be found. I had thought changing the semantics of pickVertex() as you suggested but as this is a virtual function implemented by different geometries I thought better of it. I think keeping the existing vertex in this situation is sensible, but this decision should probably be moved to the routine calling pickVertex().
With the point_vernon.laz dataset, the squares representing the points are drawn at 0.2 x 0.2 m. This is why I used a cylinder of radius 0.1m for finding points in part (a) of the algorithm - and it worked nicely (well I thought it did). For other datasets where the squares are drawn at a different size this will cause a problem and we need to make the cylinder diameter the same width as the squares so that the algorithm will select the square the user clicked on, even if it is way in front or way behind their last point of interest. We can talk more when we catch up this afternoon.

c42f:
Ah, I think I understand what you mean by squares now: the points might come out as squares if there's OpenGL driver problems.  Normally they're rendered as little spheres by the fragment shader.

gphysics:
Yes. I've just installed it on my laptop and merged your master (which was one commit ahead) with my forked master. When I run it on my laptop I get circles (or spheres with flat shading, I'm not sure which).

gphysics:
From our testing today it seems to me the speed performance is adequate - not sure that we need to improve on this (e.g. with more clever octree testing) for now. Also you could try different maximum nodes per leaf in the octree decomposition. Do we need to tidy up the mixed use of floats and doubles? I'm happy to have a crack at this (within the routines that I've touched) before you merge into the master branch.

gphysics:
Probably also should move the closestPoint search call into the octree struct.

src/pointarray.cpp
```diff
@@ -61,6 +61,71 @@ struct OctreeNode
             delete children[i];
     }
 
+/*    void print(const int indent, const bool treeWalkDown, const V3f *const P) const
```
c42f: Let's uncomment this or delete it.  I'm happy to have debug code compiled as long as it doesn't get called.  At least if it's compiled, then it'll be maintained.

c42f:
Moving the closest point search sounds reasonable.  Mixing doubles and floats doesn't bother me too much unless there's a demonstrable difference in speed.  I would just use doubles for it all if I wasn't trying to save memory on the point position arrays.

For the method of traversal, the current method certainly seems like it's fast enough.  I have a vague worry that it's not actually computing any simply defined function, since the octree traversal can terminate without finding any points.  I'd probably be happy to live with this if we sort out the issues with `m_lastVertexPickIdx` (ie, delete it and the fix the API).  I'll do that before merging unless you feel inclined to take a crack at it (I suggest changing `pickVertex` to return a bool exit code, and make the picked vertex position a pointer parameter).

gphysics:
I like your thoughts. I'll implement the above and change the API as you suggest.

gphysics:
Done. Have a look at the diff and let me know of anything you don't like.

src/pointarray.cpp
```diff
@@ -61,6 +61,48 @@ struct OctreeNode
             delete children[i];
     }
 
+    bool rayPassesNearOrThrough(
+        const V3d& rayOrigin,
+        const V3d& rayDir) const
```
c42f: Stylistic quibble: Could you put the function arguments on a single line when they fit?  There's nothing wrong with what you have here of course, other than it being a bit inconsistent with the ambient style of the codebase.

src/pointarray.cpp
```diff
@@ -61,6 +61,48 @@ struct OctreeNode
...
+        // rayOrigin lies outside of bounding sphere
+
+        const double cosA = (o2c ^ rayDir)/l;  // cosine of angle between rayDir and vector from origin to center
+
+        if(cosA < DBL_MIN)
+            return false; // rayDir points to side or behind with respect to direction from origin to center of box
+
+        const double sinA = sqrt(1 - cosA*cosA); // sine of angle between rayDir and vector from origin to center
+
+        return sinA/cosA < diagRadius/l;
```
c42f: Further stylistic quibble: there's a lot of whitespace here.  Again for consistency, I'd prefer if statements have no whitespace between them unless they form a logically distinct block in the function.

src/pointarray.cpp
```diff
@@ -61,6 +61,48 @@ struct OctreeNode
...
+        double& thisRClosest) const
+    {
+         // calls utils closestPointToRay for the points within this octree node
+         return beginIndex + closestPointToRay(
+                                 p + beginIndex,
+                                 endIndex - beginIndex,
+                                 rayOrigin,
+                                 rayDirection,
+                                 longitudinalScale,
+                                 &thisRClosest);
```
c42f: This is neater, thanks.

Again with the function arguments, I prefer compact code where possible (function arguments on one line), but readability counts so I'll leave that up to your judgment here.

For formatting consistency, if you can indent relative to the opening brace that would be helpful, thus:

```C++
return beginIndex + closestPointToRay(p + beginIndex,
                                      endIndex - beginIndex,
                                      rayOrigin,
                                      rayDirection,
                                      longitudinalScale,
                                      &thisRClosest);
```

src/view3d.cpp
```diff
@@ -535,7 +535,8 @@ Imath::V3d View3D::guessClickPosition(const QPoint& clickPos)
     // position x,y and the z of the reference position.  Take the reference to
     // be the camera rotation center since that's likely to be roughly the
     // depth the user is interested in.
-    V3d refPos = m_camera.center();
+//    V3d refPos = m_camera.center();
```
c42f: Let's remove this commented out line.  The comment above also needs some editing since it's getting out of date.  How about:

```C++
// Get new point in the projected coordinate system using the click
// position x,y and the z of a reference position.  Take the reference point
// of interest to be between the camera rotation center and the camera
// position, as a rough guess of the depth the user is interested in.
//
// This works pretty well, except when there's noise points intervening
// between the reference position and the user's actual point of interest.
```

src/view3d.cpp
```diff
@@ -551,28 +552,29 @@ Imath::V3d View3D::guessClickPosition(const QPoint& clickPos)
...
-        V3d p = m_geometries->get()[geomIdx]->pickVertex(cameraPos, pos, viewDir,
-                                                         normalScaling, &dist, &info);
-        if (dist < nearestDist)
-        {
-            newPos = p;
-            nearestDist = dist;
-            pointInfo = QString::fromStdString(info);
-        }
+        if(m_geometries->get()[geomIdx]->pickVertex(cameraPos, pos, viewDir, normalScaling,
+                                                    pickedVertex, &dist, &info))
```
c42f: For `if` statements with multi-line bodies, please insert braces.

c42f:
Thanks for fixing the API, this looks good now.  I have a few stylistic quibbles as I've written in the line notes, but I think this is just about ready.

gphysics:
Style improvements implemented.

c42f:
I fixed up something I noticed in testing (point attributes printed to the log even when no point is selected), rebased & merged.  Looks good!

gphysics:
Why does it say my branch has unmerged commits? I fixed the style and committed an hour ago...

c42f:
Ack, probably because I merged it but forgot to push from work.  Lemme have a look.

c42f:
Hmm, no it's all there on master and merged.  It's probably just because I did a rebase to clean up the branch history a bit since it was getting slightly messy.  If you rebase your branch onto 1b4541a the web interface will probably notice that it's actually been merged.

c42f:
Actually, scratch that, you'll probably get a bunch of conflicts if you try to do such a rebase.  The way to force the web interface to think it's all merged would be to do a `git reset --hard 1b4541a && git push -f`, after checking the logs that all your commits did indeed make it onto master (I'm quite sure they did).

c42f:
Note also that you'll need to be on the `closest_point_optimization` branch for the above to make sense.  Really it's no big deal - the web interface might be confused, but if you look in master you'll see it's all merged.

gphysics:
Yes it did all come across. I'll delete my branch on github. I'm keen to work on another issue/enhancement. Anything at the top of your list of priorities?

c42f:
Cool.  Anything on the issue list which isn't being actively worked on is fair game.  Some things which would be particularly useful for users are #1, #9, #12, #15, #43, #50.  Some things which involve fun graphics stuff include #8, #44.

There's also an entire epic of new development around the hierarchical point cloud stuff I mentioned the other day.  No issue exists despite the large amount of work I've already done on it, but I should make one.

PR #49, opened 2015-06-09
Added OpenSuse Install dependencies in README.rst

gphysics:


c42f:
Looks good, thanks I will merge it.

Possibly interesting side comment: The only way I used to have of figuring out a definitive list of packages was installing a new system from scratch, possibly in a chroot.  Lately I've been using docker containers for this kind of thing.  They make it super easy to get stripped down userland for any of the mainstream distros, on top of which you can quickly install packages as necessary.

PR #56, opened 2015-06-15
Animated view transforms R and T (Revisited)

nigels-com:
Couldn't resist tinkering with https://github.com/c42f/displaz/pull/54

Rebased onto master branch and modestly refactored to be frame-rate independent.

### DIFF
--- a/src/interactivecamera.h
+++ b/src/interactivecamera.h
@@ -162,7 +162,7 @@ class InteractiveCamera : public QObject
         }
         void setRotation(QQuaternion rotation)
         {
-            m_rot = rotation;
+            m_rot = rotation.normalized();
             emit viewChanged();
         }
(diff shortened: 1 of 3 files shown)


nigels-com:
Seems to me we don't need two separate timers.  This change re-purposes the existing one to trigger redraws, and moves the animation logic into paintGL.  The idea is for paintGL be called continuously for the duration of the animation.

nigels-com:
Withdrawing this pull request, for now.  (My branch now seems buggy)

c42f:
Ok, no probs.  I think keeping the discussion in one place on #54 makes sense anyway.  @gphysics has your first commit on his branch now.

nigels-com:
Sure thiing.

PR #57, opened 2015-06-16
Plot multiple point clouds with random color with python

OmriTreidel:
Additional plotting function in the python binding for plotting multiple point clouds

### DIFF
--- a/bindings/python/displaz.py
+++ b/bindings/python/displaz.py
@@ -90,6 +90,35 @@ def plot(position, plotspec='b.', color=None):
     _call_displaz(*args)
 
 
+def multi_plot(position_list):
+    """
+    Plots different points clouds with random colors
+    
+    Args:
+        position_list (list(numpy.array)): list of point clouds
+    """
+    if not isinstance(position_list, list):
+        raise TypeError("Input should be a list of points clouds")
+    file_names = []    
+    for i, position in enumerate(position_list):
+        if position.shape[1] != 3:
+            raise Exception('Position must have three columns')
+        if position.dtype != np.float64:
+            position = np.float64(position)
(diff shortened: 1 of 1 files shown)


c42f:
Cool, thanks for the pull request :-)  This looks undeniably useful, though I feel a bit unsure about having a new function for this very specific use case.  (The fact that you need a new function is basically my fault for not thinking through the binding API well at all.)

Maybe it would be better in the medium term to do something like in the C++ binding: make a point list class where it's easy to append points with arbitrary attributes, to be plotted all at once at the end.  The you could build a version of `multi_plot` as a trivial wrapper.  Let's see, how might the usage of such a thing look.

```python

position_list = whatever()

points = displaz.PointArray()
for i, position in enumerate(position_list):
    points.append(position, color=np.random.rand(3), cloud_index=i)
displaz.plot(points)
```

Not too bad I guess?  The attribute `cloud_index` is just there to illustrate how you could add additional attributes onto the point list (attributes which become available in the shader).

c42f:
Sorry for sitting on this so long.  Are you still interested in having this?

I don't want to merge what you've got in quite the current state, but I have in mind something that would be harmonious with user expectations from other plotting packages:  Just make `displaz.plot()` select a new deterministic colour each time a data set is plotted.  If the user doesn't explicitly set the colour, use the autogenerated version.

OmriTreidel:
No worries. I've made my own copy of python binding do I'm using it :) I
agree that it's not the best way to do it, it was the only way I could
think off.

c42f:
In the interests of cleaning up, I'm going to close this since I don't want to merge it as-is.  Feel free to open a new one though.

OmriTreidel:
OK, cheers

Issue #50, opened 2015-06-12
3D artifacts with Intel integrated GPUs


c42f:
Displaz doesn't seem to work well with intel graphics drivers on windows (+ possibly other platforms): points in the foreground go missing, and the depth testing is messed up with foreground points sometimes appearing behind background points.

c42f:
The missing points can be made to come back by removing the use of `gl_FragDepth` from the fragment shader.  Not sure about the depth testing yet, it's mysterious.

c42f:
@scottcarpenter FYI, I plan to stop brushing this one off ;-)

c42f:
Ok, there's at least two separate things going on here:

* The Intel drivers are somehow unhappy about `glBlitFramebuffer()`, or about the format of the incremental framebuffer itself.  It looks like depth testing is off for the incremental framebuffer.  Disabling incremental rendering "fixes" this for small datasets.
* The Intel driver seems to have a bug which causes the storage for `gl_FragDepth` and `gl_PointSize` to alias each other or otherwise interfere somehow.  Whatever it is, setting `gl_FragDepth` causes `gl_PointSize` to be corrupted.  Example shader:

```glsl
#version 130
// Copyright 2015, Christopher J. Foster and the other displaz contributors.
// Use of this code is governed by the BSD-style license found in LICENSE.txt

uniform mat4 modelViewMatrix;
uniform mat4 projectionMatrix;
uniform mat4 modelViewProjectionMatrix;

//------------------------------------------------------------------------------
#if defined(VERTEX_SHADER)

uniform float pointRadius = 0.1;   //# uiname=Point Radius; min=0.001; max=10
// Point size multiplier to get from a width in projected coordinates to the
// number of pixels across as required for gl_PointSize
uniform float pointPixelScale = 0;


in vec3 position;

void main()
{
    vec4 p = modelViewProjectionMatrix * vec4(position,1.0);
    float pointScreenSize = clamp(2*pointPixelScale*pointRadius / p.w, 0, 400);

    gl_PointSize = pointScreenSize;
    gl_Position = p;
}


//------------------------------------------------------------------------------
#elif defined(FRAGMENT_SHADER)

out vec4 fragColor;

void main()
{
    gl_FragDepth = gl_FragCoord.z;
    fragColor = vec4(vec3(gl_PointCoord.y), 1);
}

#endif

```

nigels-com:
[...]

PR #67, opened 2015-07-15
Resolve Intel GPU rendering artifacts

nigels-com:
Resolve Issue #50 

-----------------

https://github.com/c42f/displaz/issues/50

Qt 4.8 uses the multisampling codepath for samples==1,
which the Intel OpenGL driver does not seem to like.
Forcing samples=0 as a workaround.


c42f:
This is a huge improvement, thanks!

PR #68, opened 2015-07-15
Resolve debug-mode Qt assertion converning invalid range for Geometry…

nigels-com:
I came across an assertion with a debug-enabled Qt 4.8
The value of "last" was being passed -1 for empty container.


c42f:
Good catch, rebased & pushed.

c42f:
Hmm, I wish github would detect rebase + push as an accepted pull request.

PR #69, opened 2015-07-15
Intel HD 3000 OpenGL driver fails to attach depth buffer to FBO, add …

nigels-com:
Check for failure to make an FBO attachment and log an error message, rather than continuing silently.  A workaround for the Intel multi-sampling is in a separate pull request.

### DIFF
--- a/src/view3d.cpp
+++ b/src/view3d.cpp
@@ -228,8 +228,11 @@ std::unique_ptr<QGLFramebufferObject> View3D::allocIncrementalFramebuffer(int w,
     // for samples==1, so work around it by forcing 0, if possible
     fboFmt.setSamples(fmt.samples() > 1 ? fmt.samples() : 0);
     //fboFmt.setTextureTarget();
-    return std::unique_ptr<QGLFramebufferObject>(
-        new QGLFramebufferObject(w, h, fboFmt));
+    std::unique_ptr<QGLFramebufferObject> fbo;
+    fbo.reset(new QGLFramebufferObject(w, h, fboFmt));
+    if (fbo.get() && fbo->attachment()==QGLFramebufferObject::NoAttachment)
+        g_logger.error("%s", "Failed to attach FBO depth buffer.");
+    return fbo;
 }
 
 


src/view3d.cpp
```diff
@@ -226,8 +226,14 @@ std::unique_ptr<QGLFramebufferObject> View3D::allocIncrementalFramebuffer(int w,
     fboFmt.setAttachment(QGLFramebufferObject::Depth);
     fboFmt.setSamples(fmt.samples());
     //fboFmt.setTextureTarget();
-    return std::unique_ptr<QGLFramebufferObject>(
-        new QGLFramebufferObject(w, h, fboFmt));
+    QGLFramebufferObject* fbo = new QGLFramebufferObject(w, h, fboFmt);
```
c42f: Nitpick: I generally  prefer to wrap in `unique_ptr` immediately at allocation time to make early function exits safe from memory leaks (including via unexpected exceptions).

src/view3d.cpp
```diff
@@ -226,8 +226,14 @@ std::unique_ptr<QGLFramebufferObject> View3D::allocIncrementalFramebuffer(int w,
...
+    switch (fbo->attachment())
+    {
+        case QGLFramebufferObject::NoAttachment:
+            g_logger.error("%s", "Failed to attach FBO depth buffer.");
+            break;
```
c42f: Nitpick 2: the switch seems a bit unnecessary here, vs just using an `if`?

src/view3d.cpp
nigels-com: Fair enough.  Can we assume C++11 and use auto here?

src/view3d.cpp
nigels-com: Yeah, I considered logging the other possible attachment types, but decided against it in the end.
I don't mind either way.

nigels-com:
Refactored to address review nit-picks.

c42f:
Looks good thanks I thought it was probably a multi case switch which had been refactored down to one case.

Since MSVC is the rate limiting step for C++11, am currently assuming any C++11 features which are compatible with roughly VS 2012 fair game.  We could up that to VS 2013 I'll add a development FAQ section to the README.

src/view3d.cpp
```diff
@@ -228,8 +228,11 @@ std::unique_ptr<QGLFramebufferObject> View3D::allocIncrementalFramebuffer(int w,
     // for samples==1, so work around it by forcing 0, if possible
     fboFmt.setSamples(fmt.samples() > 1 ? fmt.samples() : 0);
     //fboFmt.setTextureTarget();
-    return std::unique_ptr<QGLFramebufferObject>(
-        new QGLFramebufferObject(w, h, fboFmt));
+    std::unique_ptr<QGLFramebufferObject> fbo;
+    fbo.reset(new QGLFramebufferObject(w, h, fboFmt));
+    if (fbo.get() && fbo->attachment()==QGLFramebufferObject::NoAttachment)
```
c42f: BTW, I don't know how `fbo.get()` could ever return null so this seems overly defensive.  The only way for the constructor to error out is to throw.  If it can't be allocated it'll be a `std::bad_alloc`, otherwise whatever Qt-defined exception.  (Does qt even use exceptions?  AFAIK not.)

c42f: I merged before I noticed this anyway, and don't want to bother you for another trivial change in any case.

c42f:
Rebased & pushed.

PR #72, opened 2015-07-24
Initial port to QT5

dedalusj:
Initial port of display to QT5. 

I got to the point where I could compile and run it on my Mac even though the drawing is broken. 

All the controls seems to be working but it doesn't draw any of the points or geometric figures. I suspect this is due to the way QT5 handles the OpenGL context now but I'm not 100% sure.

This was pretty much a day effort so no guarantee of niceness :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,13 +8,11 @@ set(CMAKE_BUILD_TYPE "Release" CACHE STRING
     "Choose the type of build, options are: None(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) Debug Release RelWithDebInfo MinSizeRel."
 )
 
-if(NOT WIN32)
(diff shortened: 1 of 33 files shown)


nigels-com:
Is it Qt5 requiring the core context, or displaz requiring OpenGL 3?  Is the displaz roadmap for core-only support?

c42f:
Thanks for the path!  My impression is that Qt 4 is more or less end of life now so a port to Qt5 is inevitable.  Just to check - are you planning on fiddling with this any more or should someone else take over?

@nigels-com - I'm trying to write new code against the core profile.  I think some of the features used by displaz are OpenGL 3 (some shader features IIRC), but I don't actually know the minimum viable OpenGL version.

c42f:
Superseded by #79

PR #73, opened 2015-08-20
Miscellaneous minor fixups

nigels-com:


src/mainwindow.cpp
```diff
@@ -436,8 +439,18 @@ void PointViewerMainWindow::handleMessage(QByteArray message)
...
 void PointViewerMainWindow::keyReleaseEvent(QKeyEvent* event)
 {
-    if(event->key() == Qt::Key_Escape)
-        close();
+    switch(event->key())
+    {
+        case Qt::Key_Escape:
+            close();
+            break;
+        case Qt::Key_F11:
```
c42f: I think this might be better as a global keyboard shortcut, as set using a `QAction` in the view menu?  That way it's a lot more discoverable, since the keyboard shortcut will show up on the menu.

I just noticed in the docs that there's also `setWindowState(Qt::WindowFullScreen)` - if you want a truer full screen that might make more sense too?

src/las_io.cpp
```diff
@@ -287,26 +287,26 @@ bool PointArray::loadLas(QString fileName, size_t maxPointCount,
...
         // but that will be fixed during reordering.
         for (size_t i = 0; i < fields.size(); ++i)
             fields[i].size = npoints;
-        totPoints = readCount;
+        totalPoints = readCount;
     }
-    if (totPoints > 0)
-        centroid = (1.0/totPoints)*Psum;
+    if (totalPoints > 0)
+        centroid = (1.0/totalPoints)*Psum;
```
c42f: Thanks, I like these changes.

src/fileloader.h
```diff
@@ -105,6 +108,8 @@ class FileLoader : public QObject
                     this, SIGNAL(loadProgress(int)));
             connect(geom.get(), SIGNAL(loadStepStarted(QString)),
                     this, SIGNAL(loadStepStarted(QString)));
+            connect(geom.get(), SIGNAL(resetProgress()),
+                    this, SIGNAL(resetProgress()));
```
c42f: Thanks for having a look at this.  I left it in a bit of a mess after some refactoring a while back which really improved some other things but messed up the loading bar.

It would be simpler just to emit `resetProgress` directly from `FileLoader` after invoking `Geometry::loadFile`, rather than involving the Geometry class at all.  If you think it makes a lot more sense to have the event emitted by `Geometry`, I'd rather call it `Geometry::loadFinished` or something.

In the process of looking at this I've noticed I never disconnected the signals from `geom` after the load.  This is a hangover from the way it used to work where the FileLoader was deleted after loading a group of files.  The `FileLoader` is now persistent so we should probably disconnect the signals explicitly to avoid the number of connections growing without bound as files are loaded and unloaded.

src/mainwindow.cpp
nigels-com: Turned it into a self-contained QAction.

src/fileloader.h
nigels-com: Agreed.  Updated.

nigels-com: I added the geometry signal cleanup too.

nigels-com:
Ready for a second look.

src/fileloader.h
```diff
@@ -131,6 +134,12 @@ class FileLoader : public QObject
             {
                 g_logger.error("Error loading %s: %s", loadInfo.filePath, e.what());
             }
+
+            if (geom.get())
+                geom->disconnect();
```
c42f: This doesn't do anything: if geom could be null, we'd already have crashed by now (it's used unconditionally further up).

src/qtmaximise.h
```diff
@@ -0,0 +1,35 @@
...
+
+    public slots:
+        void handleTriggered(bool checked)
+        {
+            const bool maximised = m_mainWindow->windowState()&Qt::WindowMaximized;
+            if (maximised)
+                m_mainWindow->showNormal();
+            else
+                m_mainWindow->showMaximized();
+        }
```
c42f: IMO this class isn't very idiomatic Qt: it tightly couples two components (A QActionMaximize and QMainWindow) together, rather than loosely coupling things with signals and slots.

Here's the way I'd do it:
* I'd add a new slot `PointViewerMainWindow::toggleMaximized()`
* I'd create a normal `QAction`, and connect its `triggered` signal to `toggleMaximized`

That way I think we can avoid having an entire new class for this.

src/fileloader.h
nigels-com: Seemed to me that geometry loading could (in principle) fail so severely it would make sense to nuke the object.  I don't mind being defensive, or not defensive.

src/qtmaximise.h
nigels-com: Ah, but Qt is object-oriented.  And QActionMaximize is legitimately specific to QMainWindow.
What bothers me more is using Australian english for a Maximise class calling Maximize.

src/fileloader.h
c42f: The point is, bool(geom.get()) *always* returns true since geom cannot be null at this point.  Therefore, `geom->disconnect()` should be unconditional since the if statement only serves to bloat up the code for no reason.

src/qtmaximise.h
c42f: I'm entirely happy to use American spelling for consistency.

It's true that this is specific to `QMainWindow`, fair enough.  My underlying quibble is that creating a new class for this seems like a lot of boilerplate for a tiny piece of functionality.  The alternative is much less code:

Two additional lines in `PointViewerMainWindow::PointViewerMainWindow`:
```C++
QAction* maximizeAct = viewMenu->addAction("Maximize");
maximizeAct->setShortcut(Qt::Key_F11);
connect(maximizeAct, SIGNAL(triggered(bool)), this, SLOT(toggleMaximized()));
```

Seven lines to define the new slot, plus one in the header.
```C++
PointViewerMainWindow::toggleMaximized()
{
    if (windowState() & Qt::WindowMaximized)
        showNormal();
    else
        showMaximized();
}
```

So a total of 10 new lines for this functionality, vs 30 lines and a new header file.  What you've got is better factored, but I'm not really convinced the tradeoff is worth it.

c42f: Oh, by the way - if you manage to convince me on this one, let's rename qtmaximize.h to QMaximizeAction.h - I've started switching over to use header names which reflect the main class therein.  A while ago I came to the conclusion that my old convention of "use lowercase names which might not exactly reflect the class name" was just pointlessly confusing.

nigels-com: Concerning QActionMaximise: Not sure it's worth quibbling over. In the past I've had various QAction collections for populating right-click context menus, rather than wiring those directly into QMainWindow by default, which can wind up being monolithic.  I would have liked QActionMaximise to be more context-sensitive - such as disabling itself when the window is already maximised via the window manager.  But I retreated on that when I ran into complications.  It would be easier to tackle that in isolation to the main window.

A broader issue this raises in my mind is possibly separating the core "model" and "rendering" from the broader GUI stuff.  Aside from other branches and forks the code seems due for a bit of a shuffle?

src/fileloader.h
nigels-com: Resolved.

src/qtmaximise.h
c42f: Yes, I think the code could do with a good reshuffle.  I opened an issue rather than discuss details here - see #74 for initial thoughts.

Regarding the `QActionMaximise` class - implementation wise I'm willing to go with it and see what it feels like.  However, I have one final objection on the design front which I only just realized was hovering in the back of my mind bugging me: window managers usually have shortcut keys for this kind of thing already - why not just learn that?  It's a win-win solution: it'll work for all your windows, and displaz will have less code.  Even windows has such a shortcut: `ctrl+win+↑`

nigels-com: win+↑ does something else on Windows 10, apparently.

c42f: Regardless of the details of exactly which keyboard shortcut is chosen on a given system, is this not a window manager concern?

c42f: (I hope I'm not being too obtuse about this, but personally I use a window manager which actually manages windows, so this shortcut doesn't seem to add any real value.  It seems reasonable that people could learn to use the standard keybindings provided by their window managers, whatever those actually are.)

c42f: I could certainly be wrong about win+↑, I just looked it up on the web and assumed it actually worked ;-)

nigels-com: So it turns out that the gnome default for toggling window maximization is ALT-F10.  Which I've happily remapped to F11.  But having solved that, displaz is not pure - the ESC key for closing the whole application has bitten me more times than I care to mention.  Conventionally, close a dialog with ESC, but not the entire application.

c42f: You're right, let's remove the escape shortcut.

nigels-com:
Refinements made, as discussed.  Third time lucky?

c42f:
Cheers, thanks for putting up with all that!  Please do continue questioning the design - there's plenty of areas where the existing design isn't really clean or consistent.  `PointViewerMainWindow` is definitely not an example of elegantly factored code, but if it's straightforward and short, perhaps that's ok...  It's the classic worse is better type tradeoff ;-)

PR #75, opened 2015-09-21
Plot lines in Julia

JoshChristie:


c42f:
Great, thanks for the patch, I've wanted to be able to do this for a while.  As you can probably see, my interface for the julia bindings isn't very well thought out...

bindings/julia/Displaz.jl
```diff
@@ -50,15 +49,54 @@ function write_ply_points(fileName, nvertices, fields)
             for i = 1:size(value,2)
                 propName = ply_property_name(semantic, i)
                 write(fid, "property $typeName $propName\n")
-            end
+        	end
         end
         write(fid, "end_header\n")
         for (_,value) in convertedFields
-            write(fid, value')
+			write(fid, value')
```
c42f: Formatting: can you replace tabs -> spaces for formatting consistency.

bindings/julia/Displaz.jl
```diff
@@ -50,15 +49,54 @@ function write_ply_points(fileName, nvertices, fields)
...
+function write_ply_lines(fileName, position, color)
+	vertexValid = ~all(isnan(position),2)
+	lineStarts = find(x -> x == 1, diff([false;vertexValid],1))
+	nvalidvertices = sum(x -> x == true, vertexValid)
+
+	# Create and write to ply file
+	fid = open(fileName, "w")
+	write(fid, "ply\n")
+	write(fid, "format binary_little_endian 1.0\n")
+	write(fid,"element vertex $nvalidvertices\n")
```
c42f: formatting quibble - inconsistent spacing

bindings/julia/Displaz.jl
```diff
@@ -50,15 +49,54 @@ function write_ply_points(fileName, nvertices, fields)
...
+    nvertices = size(position,1)
+    realStart = 0
+    for i = lineStarts'
+        j = i
+        while j <= nvertices && vertexValid[j]
+            j = j + 1
+        end
+        lineLen = j-i
+		write(fid,int32(lineLen))
+        write(fid,int32(realStart:realStart+lineLen-1))
```
c42f: `int32` is deprecated in 0.4, it's now `Int32`.  You can also use `collect()` to create an array from any iterable such as the range `realStart:realStart+lineLen-1`:

```julia
write(fid, Int32(lineLen))
write(fid, collect(Int32, realStart:realStart+lineLen-1))
```

bindings/julia/Displaz.jl
```diff
@@ -50,15 +49,54 @@ function write_ply_points(fileName, nvertices, fields)
...
+	write(fid,"property double y\n")
+	write(fid,"property double z\n")
+	write(fid,"element color $nvalidvertices\n")
+	write(fid,"property float r\n")
+	write(fid,"property float g\n")
+	write(fid,"property float b\n")
+	write(fid,"element edge $(length(lineStarts))\n")
+	write(fid,"property list int int vertex_index\n")
+	write(fid,"end_header\n")
+	idx = find(x-> x== true,vertexValid)
```
c42f: You should be able to use `find(vertexValid)` here, and delete the `x->x==true`

bindings/julia/Displaz.jl
```diff
@@ -50,15 +49,54 @@ function write_ply_points(fileName, nvertices, fields)
...
         write(fid, "end_header\n")
         for (_,value) in convertedFields
-            write(fid, value')
+			write(fid, value')
         end
     end
 end
 
+function write_ply_lines(fileName, position, color)
+	vertexValid = ~all(isnan(position),2)
```
c42f: It's a bit unfortunate to inherit the matlab hack of separating the position array using nans.  (The only reason the matlab bindings do it is because it's a relatively fast way to plot a lot of disconnected lines using the builtin matlab plot command.)

Ideally I'd like to settle on a more sensible convention (for instance, passing the vector of lineStarts explicitly)

bindings/julia/Displaz.jl
```diff
@@ -50,15 +49,54 @@ function write_ply_points(fileName, nvertices, fields)
...
+	lineStarts = find(x -> x == 1, diff([false;vertexValid],1))
```
c42f: The julia naming convention (which I try to follow, though I can see there's some slipups elsewhere) is to use `underscore_case`, or `squishedcase` for short names.  Can you change the variable names to follow that convention for consistency?

bindings/julia/Displaz.jl
```diff
@@ -107,25 +155,27 @@ function plot(position; color=[1 1 1], markersize=[0.1], markershape=[0])
...
     #fileName = "_julia_tmp.ply"
     fileName = tempname()*".ply"
-    write_ply_points(fileName, nvertices, (
-                     (:position, vector_semantic, position),
-                     (:color, color_semantic, color),
-                     (:markersize, array_semantic, markersize),
-                     (:markershape, array_semantic, markershape),
-                     ))
-    hold = _hold ? "-add" : []
+	if isequal(markershape[1],5) # Plot lines
```
c42f: This could probably just be

`if '-' in markershape`

Then it'd also work when markershape is a `String` of some sort. and be less obscure as well.

c42f:
Hmm, looking through this really suggests I need to think more carefully about a good plotting API.  `plot()` just doesn't really have a convenient interface, with the keywords being harder to use than a matlab-style plotspec string.  That's probably not something for this pull request though.

bindings/julia/Displaz.jl
JoshChristie: Will do :)

bindings/julia/Displaz.jl
JoshChristie: okay will try this when i get some time

bindings/julia/Displaz.jl
JoshChristie: I found this quite annoying. thanks for the tip

bindings/julia/Displaz.jl
JoshChristie: okay

bindings/julia/Displaz.jl
JoshChristie: I see thanks :)

bindings/julia/Displaz.jl
JoshChristie: Okay

bindings/julia/Displaz.jl
c42f: maybe we live with this for now, and try to figure out a better way of passing multiple line segments as a separate issue.

Did you test the nan functionality?  When I tried the following it didn't result in two line segments like I expected:
```julia
julia> P = [1.0 1 0;
            2 1 0;
       3 1 0;
       NaN NaN NaN;
       1 2 0;
       2 2 0;
       3 2 0]
7x3 Array{Float64,2}:
   1.0    1.0    0.0
   2.0    1.0    0.0
   3.0    1.0    0.0
 NaN    NaN    NaN  
   1.0    2.0    0.0
   2.0    2.0    0.0
   3.0    2.0    0.0

julia> using Displaz
julia> Displaz.plot(P, markershape='-')
```

c42f: On second thoughts, let's kill off the NaNs now.  I suggest that we simply pass the equivalent of lineStarts as a positional argument in the interface - let's call it `linebreak` (or some better name?).  In the plot function it could be a keyword argument, with a default of `[1]`.  Example usage via plot:

```julia
# Plot two separate line segments, with the index of the first starting at 1,
# and the index of the second starting at 11.  Internally this should result
# in the ranges `1:10` and `11:20` for the two line segments.
Displaz.plot(randn(20,3), markershape='-', linebreak=[1,11])
```

bindings/julia/Displaz.jl
```diff
@@ -79,14 +117,24 @@ interpret_color(s::String) = length(s) == 1 ? interpret_color(s[1]) : error("Unk
...
+interpret_shape(markershape) = markershape
+interpret_shape(c::Char) = shape_names[c]
+
 # True if plots are to be plotted over the previous data, false to clear before
 # plotting new data sets.
 _hold = false
 
-
 # Basic 3D plotting function for points
 function plot(position; color=[1 1 1], markersize=[0.1], markershape=[0])
```
JoshChristie: Can we rename this function as it conflicts with other plotting functions? Any preferences @c42f ?

c42f: Sure, I'm happy to rename some things.  As I said somewhere else, this whole API needs a rethink...  The underlying issue is that there's not a well established convention for plotting APIs in julia.  Or at least there wasn't when I started this.

Since then there's been a lot of development.  One of the most interesting packages from an API point of view is an attempt to unify plotting here:

https://github.com/tbreloff/Plots.jl

I really like the convention from Plots.jl of doing away with `hold()` and just using `plot()` and `plot!()`.

As far as naming goes, `plot3()` could make sense I suppose.

JoshChristie:
Specifying linebreak as an integer N will divide the points in line segments of size N. 
``` 
P = [1.0 1 0;
      2 1 0;
      3 1 0;
      1 2 0;
      2 2 0;
      3 2 0]
plot(P, markershape = '-') # Plot all points
plot(P, markershape = '-', linebreak = [1,3]) # Plot 2 lines
plot(P, markershape = '-', linebreak = 2)      # Plot 3 lines (i.e. Each line contains 2 points)
```

bindings/julia/Displaz.jl
```diff
@@ -37,28 +37,71 @@ function ply_property_name(semantic, idx)
...
+        property float r
+        property float g
+        property float b
+        element edge $(length(linebreak))
+        property list int int vertex_index
+        end_header
+        """
+    )
+
+    write(fid,convert(Array{Float64,2},position'))
```
c42f: What's the `convert` do here?

c42f: I'm going to assume this is something to do with position sometimes being an array of the wrong type (eg, an integer type).  I'll leave it in for now since I'd like to merge this.  We can probably do something more consistent with the types before passing to `write_ply_*` later.

c42f:
Squashed + merged

Issue #76, opened 2015-10-04
Automated build testing (on multiple platforms)


chrisidefix:
I suggest adding an (online) build test setup (for example using https://travis-ci.org) to the project. `travis` also allows testing on different platforms (linux & os x right now, but windows is also in the making). Since you mentioned, you may not have hardware access to an os x machine, this could be one way to check if the code at least compiles and runs there, too. It also won't hurt having regular build tests done.

c42f:
Definitely.  I've just started using travis for another project of mine and it seems to be pretty nice.

c42f:
see also #66

chrisidefix:
One more thing - I understand `os x` support probably isn't very high on your priority list, but if you ever do think about testing (any of) your projects there - you can actually create Mac VM's (with VirtualBox), too.  Example: http://lifehacker.com/5938332/how-to-run-mac-os-x-on-any-windows-pc-using-virtualbox

nigels-com:
I took a look at building displaz in Travis.  However, there is some kind of cmake bug concerning FindIlmbase.cmake version detection, which didn't seem to have an obvious workaround.  When cmake fails, it does tend to be an epic fail...

```
$ cmake ..
-- The C compiler identification is GNU
-- The CXX compiler identification is GNU
-- Check for working C compiler: /usr/bin/gcc
-- Check for working C compiler: /usr/bin/gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/g++
-- Check for working CXX compiler: /usr/bin/g++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
CMake Error at /usr/share/cmake-2.8/Modules/FindPackageHandleStandardArgs.cmake:91 (MESSAGE):
  FOUND_VAR (missing: REQUIRED_VARS VERSION_VAR) (Required is at least
  version "1.0.1")
Call Stack (most recent call first):
  /usr/share/cmake-2.8/Modules/FindPackageHandleStandardArgs.cmake:252 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindIlmbase.cmake:22 (find_package_handle_standard_args)
  CMakeLists.txt:50 (find_package)
```

c42f:
[...]

PR #77, opened 2015-10-05
Initial https://travis-ci.org/ configuration

nigels-com:
Regarding https://github.com/c42f/displaz/issues/76

I managed to get displaz building on Travis with a backported cmake that is slightly newer than the one for Ubuntu 12.04.

https://travis-ci.org/nigels-com/displaz


c42f:
Thanks, I think this will be handy.

PR #78, opened 2015-10-05
Update to travis ci to build OS X version as well.

chrisidefix:
As discussed in issue #76 I added support for OS X build testing on travis (failures under OS X are allowed though). I also tested a version that actually compiles CMake 2.8.12.2, but this makes the whole build process take much longer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,10 +35,10 @@ script:
     - cd ..
 
 install:
-    - sudo apt-get install wget git g++ cmake libqt4-dev libqt4-opengl-dev python-docutils
-    - wget https://launchpad.net/~kalakris/+archive/ubuntu/cmake/+build/4892863/+files/cmake_2.8.11.2-1ubuntu2~precise1_amd64.deb
-    - wget https://launchpad.net/~kalakris/+archive/ubuntu/cmake/+files/cmake-data_2.8.11.2-1ubuntu2~precise1_all.deb
-    - sudo dpkg -i cmake_2.8.11.2-1ubuntu2~precise1_amd64.deb cmake-data_2.8.11.2-1ubuntu2~precise1_all.deb
+    - if [ "$TRAVIS_OS_NAME" == "linux" ]; then sudo apt-get install wget git g++ cmake libqt4-dev libqt4-opengl-dev python-docutils; fi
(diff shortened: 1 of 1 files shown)


c42f:
Thanks for the patch!   Yes, the cmake issue is a pain.  Ultimately it's probably best to work around the brokenness by modifying cmake/FindIlmBase.cmake rather than messing with the CI build.  I've built version 2.8.7 locally and can reproduce the error, so shouldn't be too hard I hope.

.travis.yml
```diff
@@ -1,22 +1,52 @@
...
+  - osx
+
+compiler:
+  - gcc
+  - clang
+
+branches:
+  only:
+    - master
+    - travis-osx
```
c42f: How about just building all branches?  Otherwise future PRs won't get CI coverage.

.travis.yml
```diff
@@ -1,22 +1,52 @@
...
+
+before_script:
+  - if [ "$TRAVIS_OS_NAME" == "osx" ]; then brew update; fi
+  - if [ "$TRAVIS_OS_NAME" == "osx" ]; then brew install git qt4; fi
```
c42f: I must say I'm confused by the distinction between `before_script`, `install`, and `before_install`.  I'm quite new to Travis and the docs did not really enlighten me.  Regardless, shouldn't both the linux and OSX package manager commands go in the same section?

.travis.yml
```diff
@@ -1,22 +1,52 @@
...
+    - cmake ../thirdparty/external
+    - make -j4
+    - cd ..
+    # Build displaz
+    #- pwd
+    - mkdir build
+    - cd build
+    - cmake ..
+    - make -j4
+    - cd ..
```
c42f: Nitpick: if we're going to reindent things, let's make the file self-consistent.  (Ok, I know it wasn't consistent before :) )

.travis.yml
chrisidefix: Sure. I personally prefer to only build (and get notifications) regarding the most stable branch, but it depends a bit on which kind of project it is.

c42f:
Huh, I completely failed to notice that you'd cleaned up a bunch of things here, my apologies - I would have looked at it again if I'd realized.  It looks like github doesn't produce an email notification when a new patch is added to a PR.

c42f:
Looks good - I squashed the commits for simplicity since there was a bit of churn.  Pushed to master now, thanks :)

PR #79, opened 2015-10-17
Support Mac OS X CoreProfile OpenGL 3.2 (incl. Qt5 upgrade)

chrisidefix:
If you find the time, please review and any feedback or request is welcome, of course.

**What changed?**
 1. By default `displaz` is now looking for **`Qt5`**.
 1.1. *Non* of the *recent* `QOpenGL*` classes have been used intentionally, to allow support of earlier `Qt5` versions on systems, where it would be more difficult to obtain the latest release.
 1.2. The option `DISPLAZ_USE_QT4` was added to `CMake` (`FALSE` by default) and has to be selected for systems, where you want to build against `Qt4`.

 2. The **OpenGL** code was *mostly* restructured where necessary to run on a *Core Profile context 3.2*, which is the only one available for Mac OS X (OpenGL 3.0 or 3.1 can not be used and there is no support for a *Compatibility Profile*).

 2.1. **GLSL Shaders** have been added for any render step, including the "cursor" and "axes" overlays. 
 2.2. Most of the vertex/index data is now setup in **`VAO`/`VBO`** fashion through the `initializeGL()` method and then rendered with `glDrawArrays`/`glDrawElements` from `paintGL()`.

 3. I wouldn't recommend running `displaz` with `Qt4` on Mac OS X and there really is no need to do this. However, I did add some `Objective-C` code that is supposed to create a proper window with the right `CoreContext 3.2` for `QGLWidget`, in case you are wondering why this is there. For this to work `displaz` also needs to link against Apple specific frameworks (`Cocoa` and `Foundation`).

**What remains?**
 1. There is likely work left to be done for `HCloudView`. I deactivated the (unused) code in `drawBounds` for example. The `draw()` method likely needs to be adjusted, but I didn't have any dataset that I could have tested.
 2. `CoreProfile 3.2` is now set as the default `QGLFormat` inside `guimain.cpp` and it should be decided, if - for non-Mac systems - this should be something else.
 3. Some more clean-up in the code from my side.

PS: Feel free to squash the commits and ask me anything!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,11 @@ project(displaz)
 
 cmake_minimum_required(VERSION 2.8)
 
+if (WIN32)
+    # cmake >= 2.8.10 needs this for Qt5 on Windows
+    cmake_policy(SET CMP0020 NEW)
+endif()
+
 option(DISPLAZ_STATIC "Build against static libraries" FALSE)
 option(DISPLAZ_USE_LAS "Build with support for reading las files" TRUE)
 option(DISPLAZ_USE_PDAL "Use PDAL for reading las files" FALSE)
@@ -84,10 +89,6 @@ find_package(OpenGL REQUIRED)
 if (DISPLAZ_USE_QT4)
     find_package(Qt4 4.7 COMPONENTS QtCore QtGui QtNetwork QtOpenGL REQUIRED)
     add_definitions(-DDISPLAZ_USE_QT4)
-
-    if(NOT WIN32)
-        set(CMAKE_CXX_FLAGS "-Wall -std=c++0x" CACHE STRING "Flags used by the compiler during all build types.")
-    endif()
 else ()
     find_package(Qt5Core REQUIRED)
     find_package(Qt5Gui REQUIRED)
@@ -99,13 +100,11 @@ else ()
     add_definitions(${Qt5Network_DEFINITIONS})
     add_definitions(${Qt5Widgets_DEFINITIONS})
 
-    if(WIN32)
-        # cmake 3.3 needs this for Qt5 on Windows
-        cmake_policy(SET CMP0020 NEW)
-    else()
-        set(CMAKE_CXX_FLAGS "-Wall -std=c++0x ${Qt5Core_EXECUTABLE_COMPILE_FLAGS} ${Qt5Gui_EXECUTABLE_COMPILE_FLAGS} ${Qt5Network_EXECUTABLE_COMPILE_FLAGS} ${Qt5OpenGL_EXECUTABLE_COMPILE_FLAGS} ${Qt5Widgets_EXECUTABLE_COMPILE_FLAGS}" CACHE STRING "Flags used by the compiler during all build types.")
+    if (NOT WIN32)
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${Qt5Core_EXECUTABLE_COMPILE_FLAGS} ${Qt5Gui_EXECUTABLE_COMPILE_FLAGS} ${Qt5Network_EXECUTABLE_COMPILE_FLAGS} ${Qt5OpenGL_EXECUTABLE_COMPILE_FLAGS} ${Qt5Widgets_EXECUTABLE_COMPILE_FLAGS}")
     endif()
 endif()
+
 
 include_directories(${ILMBASE_INCLUDE_DIRS})
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -167,12 +167,12 @@ if (DISPLAZ_USE_QT4)
     )
 else()
     target_link_libraries(displaz
-        ${Qt5Core_LIBRARIES} ${Qt5Network_LIBRARIES} ${Qt5Widgets_LIBRARIES}
+        Qt5::Core Qt5::Network Qt5::Widgets
         ${ILMBASE_LIBRARIES}
     )
     target_link_libraries(displaz-gui
-        ${Qt5Core_LIBRARIES} ${Qt5Gui_LIBRARIES} ${Qt5OpenGL_LIBRARIES}
-        ${Qt5Network_LIBRARIES} ${Qt5Widgets_LIBRARIES}
+        Qt5::Core Qt5::Gui Qt5::OpenGL
+        Qt5::Network Qt5::Widgets
         ${OPENGL_gl_LIBRARY} ${GLEW_LIBRARIES}
         ${ILMBASE_LIBRARIES}
     )
@@ -182,9 +182,9 @@ else()
 endif()
 
 if(APPLE)
-target_link_libraries(displaz-gui
-    ${FOUNDATION_LIBRARY} ${COCOA_LIBRARY}
-)
+    target_link_libraries(displaz-gui
+        ${FOUNDATION_LIBRARY} ${COCOA_LIBRARY}
+    )
 endif()
 
 if (DISPLAZ_BUILD_DVOX)
@@ -192,7 +192,7 @@ if (DISPLAZ_BUILD_DVOX)
     if (DISPLAZ_USE_QT4)
         target_link_libraries(dvox ${QT_QTCORE_LIBRARY})
     else()
-        target_link_libraries(dvox ${Qt5Core_LIBRARIES})
+        target_link_libraries(dvox Qt5::Core)
     endif()
 endif()
 
@@ -212,12 +212,6 @@ endif()
 
 
 if(WIN32)
-    # TODO: test windows version - is this still required ?
-    if(DISPLAZ_USE_QT4)
-        target_link_libraries(displaz-gui
-            ${QT_QTMAIN_LIBRARY}
-        )
-    endif()
     if(DISPLAZ_STATIC)
         # These extra libraries are needed on windows when linking against a
         # static Qt-4.8.? which has been built with the default options.
@@ -243,9 +237,11 @@ if(WIN32)
             fixup_bundle(\"\${CMAKE_INSTALL_PREFIX}/${DISPLAZ_BIN_DIR}/displaz${CMAKE_EXECUTABLE_SUFFIX}\" \"\" \"${QT_LIBRARY_DIR}\")
         ")
     else()
+        get_target_property(QtCore_location Qt5::Core LOCATION)
+        get_filename_component(QtCore_dir "${QtCore_location}" PATH)
         install(CODE "
             include(BundleUtilities)
-            fixup_bundle(\"\${CMAKE_INSTALL_PREFIX}/${DISPLAZ_BIN_DIR}/displaz${CMAKE_EXECUTABLE_SUFFIX}\" \"\" \"${Qt5Core_LIBRARY_DIR}\")
+            fixup_bundle(\"\${CMAKE_INSTALL_PREFIX}/${DISPLAZ_BIN_DIR}/displaz${CMAKE_EXECUTABLE_SUFFIX}\" \"\" \"${QtCore_dir}\")
         ")
     endif()
 endif()
@@ -274,7 +270,7 @@ if (DISPLAZ_USE_TESTS)
     if (DISPLAZ_USE_QT4)
         target_link_libraries(InterProcessLock_test ${QT_QTCORE_LIBRARY})
     else ()
-        target_link_libraries(InterProcessLock_test ${Qt5Core_LIBRARIES})
+        target_link_libraries(InterProcessLock_test Qt5::Core)
     endif ()
     add_test(NAME InterProcessLock_test COMMAND InterProcessLock_test master)
 endif()


nigels-com:
I'm not seeing any points being rendered on Linux/NVIDIA 3.2.0 context.  But it looks promising that it builds and runs without much fuss.  I did need to add a ` -fPIC` flag manually to make Qt 5 happy.

nigels-com:
Build seems to be broken on Linux for ` cmake  -DCMAKE_BUILD_TYPE=Debug` ...

```
Scanning dependencies of target displaz
[  9%] Building CXX object src/CMakeFiles/displaz.dir/logger.cpp.o
[ 10%] Building CXX object src/CMakeFiles/displaz.dir/geomfield.cpp.o
[ 11%] Building CXX object src/CMakeFiles/displaz.dir/typespec.cpp.o
In file included from /home/nigels/dev/displaz/src/typespec.cpp:5:0:
/home/nigels/dev/displaz/src/glutil.h:27:13: warning: ‘void setUniform(GLuint, const char*, const M44d&)’ defined but not used [-Wunused-function]
 static void setUniform(GLuint prog, const char* name, const M44d& mat)
             ^
[ 12%] Building CXX object src/CMakeFiles/displaz.dir/hcloud.cpp.o
[ 14%] Building CXX object src/CMakeFiles/displaz.dir/util.cpp.o
[ 15%] Building CXX object src/CMakeFiles/displaz.dir/__/thirdparty/argparse.cpp.o
[ 16%] Building CXX object src/CMakeFiles/displaz.dir/moc_IpcChannel.cpp.o
[ 17%] Building CXX object src/CMakeFiles/displaz.dir/InterProcessLock.cpp.o
[ 18%] Building CXX object src/CMakeFiles/displaz.dir/IpcChannel.cpp.o
[ 20%] Building CXX object src/CMakeFiles/displaz.dir/main.cpp.o
Linking CXX executable ../bin/displaz
CMakeFiles/displaz.dir/typespec.cpp.o: In function `setUniform':
/home/nigels/dev/displaz/src/glutil.h:29: undefined reference to `__glewGetUniformLocation'
/home/nigels/dev/displaz/src/glutil.h:33: undefined reference to `__glewUniformMatrix4fv'
collect2: error: ld returned 1 exit status
make[2]: *** [bin/displaz] Error 1
```

c42f:
Yikes, this looks like it was a lot of work!  Thanks a heap, I'll try to take a look tonight.

c42f:
Hmm, setting fPIC manually+unconditionally seems not quite right.  The patch https://github.com/Cockatrice/Cockatrice/commit/9cfb09cd5b25a01fb142d425ba70b77b89f65a09 suggests something more like

```cmake
if (Qt5_POSITION_INDEPENDENT_CODE)
    set_target_properties(dispaz-gui PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
endif()
```

or some such.

chrisidefix:
1. The `-fPIC` flag should already be set from inside the main CMakeLists, but I'll have a look why it doesn't get picked up on linux: https://github.com/geometree/displaz/blob/osx-gl-qt5/CMakeLists.txt#L100

 2. I actually continuously build in DEBUG mode, but this looks like it is related to GLEW. Are you using the embedded GLEW? (`DISPLAZ_EMBED_GLEW = TRUE` or `FALSE`)

 3. I forgot to mention, the point rendering still has some issues, that I didn't get a chance to look at in detail.

chrisidefix:
Ok, I saw you committed some fix that I also did and will rebase my branch. Should I add build testing for `DEBUG` mode as well?

c42f:
Feel free to if you think it's relevant.  Otherwise, I think this PR has quite enough to look at!

chrisidefix:
Ok, I fixed that compile error in `Debug` mode and also extended `travis CI` to build in `Debug` & `Release` mode now. The most important thing left to do, as @nigels-com mentioned, is to fix the point rendering. (At least it doesn't crash and rendering of a single point actually works for me, but I basically haven't touched that class to make it fully compatible with the `CoreProfile`.) Second on the list would be that `HCloudView` class I think.

c42f:
It compiles and runs for me on linux (ubuntu 14.04, qt-5.2), but there's definitely a few issues as might be expected after porting to a new major version.  A grab bag of things I've seen after a quick fiddle:

* Point clouds generally don't show up.  I can see what appears to be the first point in test/standard_ply_points.ply, but the rest of the circle isn't there.
* Bounding boxes disappear for large point sets after the first frame.
* Meshes have no lighting
* The "hide the mouse when dragging" trick for numerical shader params doesn't work anymore
* The axes widget and 3D cursor look a bit different
* Some code duplication in utility shaders - maybe we can factor some of them together

I'm happy to help resolve some of these, just may not be super quick to find time.

chrisidefix:
I'll try to resolve the point cloud rendering and have a look at what is going on with the bounding boxes. I remember there was some code that used to shift the bounding box towards `bbox.min()` "for stability with large offsets", but I commented that out previously. I guess I can just reapply that fix and see if that solves them disappearing.

I'll also have a look at the lighting - I assume this is related to the `meshface.glsl` shader? Maybe the light dir or the normals aren't set properly.

Regarding the axes overlay: It's not easily possible to recreate the exact look, as `glLineWidth` can not be called with a value larger than 1.0 anymore (http://stackoverflow.com/questions/8791531/opengl-3-2-core-profile-gllinewidth). We would have to render triangle geometry or use geometry shaders for wider lines (or use a texture maybe).

Lastly, I am not sure how you used to "hide the mouse" for the shader parameter - was this some Qt based solution? I didn't see anything related to it in the code, yet.

chrisidefix:
Something else: I also added an option to render a view grid (around `0.0,0.0,0.0` for now). It is disabled by default, but there is a new option under "View > Draw Axes" if you want to try it out. (I did this mostly to see where the camera was pointing to and it's easily removed again, if it is not needed.)

nigels-com:
I tried building on Ubuntu 15.04 just now, in Qt5 mode.
Are we preferring `QOpenGLFunctions_3_2_Core` over GLEW, or doing this purely for core context support?  I do have a branch of GLEW that ought to work here, it's probably about time I merged that to the mainline...

- Nigel

```
[ 70%] Building CXX object src/CMakeFiles/displaz-gui.dir/glutil.cpp.o
/home/nstewart/dev/displaz/src/glutil.cpp: In function ‘void setUniform(GLuint, const char*, const M44d&)’:
/home/nstewart/dev/displaz/src/glutil.cpp:17:48: error: ‘glGetUniformLocation’ was not declared in this scope
     GLint loc = glGetUniformLocation(prog, name);
                                                ^
/home/nstewart/dev/displaz/src/glutil.cpp:21:43: error: ‘glUniformMatrix4fv’ was not declared in this scope
     glUniformMatrix4fv(loc, 1, GL_FALSE, &M[0][0]);
                                           ^
/home/nstewart/dev/displaz/src/glutil.cpp: In function ‘void drawSphere(const TransformState&, const V3d&, double, GLuint, const C4f&, int, int)’:
/home/nstewart/dev/displaz/src/glutil.cpp:140:28: error: ‘glUseProgram’ was not declared in this scope
     glUseProgram(shaderProg);
                            ^
/home/nstewart/dev/displaz/src/glutil.cpp:142:62: error: ‘glGetUniformLocation’ was not declared in this scope
     GLint colorLoc = glGetUniformLocation(shaderProg, "color");
                                                              ^
```

nigels-com:
I pushed my crude hacks to build Qt5 + GLEW on Linux, for reference.
https://github.com/nigels-com/displaz/commit/4abb1a87
Is the plan to use GLEW, or not?

-- Nigel

c42f:
If glew is compatible with Qt5, I'd prefer to keep using it: then we're not limited by the OpenGL APIs which the Qt people have deemed appropriate to wrap.  I found the following in qopenglcontext.h

```
#warning qopenglfunctions.h is not compatible with GLEW, GLEW defines will be undefined
#warning To use GLEW with Qt, do not include <qopengl.h> or <QOpenGLFunctions> after glew.h
```

which seems to suggest they can be used together, with care...

c42f:
@nigels-com - thanks.  Those hacks include the changes from the OpenGL core context GLEW branch do they?

nigels-com:
Yes, the GLEW I used is reported to work fine, but I have not tested it on anything non-trivial myself.  Displaz seems like a good test-bed for doing that validation.

c42f:
Sure, sounds good to me.

nigels-com:
Yeah, I rebased it onto a generic glew-1.13.0 just now.  I'd prefer the rendering to be less entangled with Qt, personally.

chrisidefix:
I would usually avoid the Qt wrapped OpenGL as Nigel suggests, which will also help to keep the actual Qt version requirement lower than 5.4, if you are still worried about that.

I updated the `point_array.cpp` to map that data into a `GL_ARRAY_BUFFER` buffer and it works for simple examples.

However, there appears to be some issue for larger point data, though. It might be worth going over this together, so I don't waste time trying to figure out how exactly the data array was setup and how it can be accessed in memory.

chrisidefix:
I completed the update to the `point_array.cpp` to map all data correctly into a single `GL_ARRAY_BUFFER` buffer and it now works also on very large (random) datasets, that I tested with.

src/corecontext.h
```diff
@@ -0,0 +1,28 @@
...
+#if defined(Q_OS_MAC)
+#if QT_VERSION < 0x050000 && QT_VERSION >= 0x040800 // if less than Qt 5.0.0
+    virtual void* chooseMacVisual(GDHandle handle) {
+        return select_3_2_mac_visual(handle, this->format().depthBufferSize()); //TODO: check if we want to enable multiSampling
+    }
+#endif
+#endif
+};
+
+#endif //DISPLAZ_CORECONTEXT_H
```
c42f: Eek, it's impressive that this works (not that I can test it) ;-)

Do you think people will want support for older Qt versions on OS X?  If not, I'd tentatively suggest we say it's version 5+ only and just delete the legacy support.

I'll certainly be using Qt5-only for the windows build as soon as I get the chance to fiddle with the build system.

src/geometry.cpp
```diff
@@ -60,3 +65,142 @@ bool Geometry::reloadFile(size_t maxVertexCount)
...
+
+    for (std::map<std::string,unsigned int>::iterator it=m_VAO.begin(); it!=m_VAO.end(); ++it)
+    {
+        GLuint vao = it->second;
+        glDeleteBuffers(1, &vao);
+    }
+
+    m_VAO.clear();
+
+}
```
c42f: Should the Geometry destructor also call `destroyBuffers()`?

src/geometry.cpp
```diff
@@ -60,3 +65,142 @@ bool Geometry::reloadFile(size_t maxVertexCount)
...
+
+void Geometry::destroyBuffers()
+{
+    // destroy any previously created buffers (in case we are re-initializing this geometry)
+    // this should avoid recreating more and more opengl buffers
+
+    //tfm::printfln("Geometry :: destroyBuffers");
+
+    for (std::map<std::string,unsigned int>::iterator it=m_VBO.begin(); it!=m_VBO.end(); ++it)
```
c42f: I think at this stage all the compilers we care about have supported `auto` for some time, so `it` can be declared `auto`.

src/pointarray.cpp
```diff
@@ -613,17 +685,58 @@ DrawCount PointArray::drawPoints(QGLShaderProgram& prog, const TransformState& t
...
+
         for (size_t i = 0, k = 0; i < m_fields.size(); k+=m_fields[i].spec.arraySize(), ++i)
         {
             const GeomField& field = m_fields[i];
             int arraySize = field.spec.arraySize();
             int vecSize = field.spec.vectorSize();
+
+            // TODO: should use a single data-array that isn't split into vertex / normal / color / etc. sections, but has interleaved data
+            // OpenGL has a stride value in glVertexAttribPointer for exactly this purpose, which should be used for better efficiency
+            // here we write only the current attribute data into this the buffer (e.g. all positions, then all colors)
```
c42f: So the suggestion is to pack all the point attributes together so that a single node worth of data can be sent with a single `glBufferSubData()`?  (That is, a packed array-of-structs style format.)  Do you know if this would have much of a positive performance boost?

It would definitely make things more painful on the CPU side, where at the moment with the struct-of-arrays format we can allocate and delete each attribute independently, and the data elements for attributes are correctly aligned for the CPU.

src/glutil.cpp
```diff
@@ -32,61 +47,91 @@ void TransformState::setUniforms(GLuint prog) const
...
+{
+    // Note: Create an orthographic matrix from input values
+    if (left == right) {
+        right = left+1.0; // TODO: should throw an error here
+    }
+    if (bottom == top) {
+        bottom = top+1.0; // TODO: should throw an error here
+    }
+    if (nearVal == farVal) {
+        nearVal = farVal+1.0; // TODO: should throw an error here
```
c42f: Sure, go ahead and throw a `DisplazError` if that makes sense.

src/shader.cpp
```diff
@@ -78,7 +78,11 @@ bool Shader::compileSourceCode(const QByteArray& src)
     {
         if (!re.exactMatch(lines[lineIdx]))
             continue;
-        QByteArray typeStr = re.cap(1).toAscii();
+#ifdef DISPLAZ_USE_QT4
+        QByteArray typeStr = re.cap(1).toAscii().constData();
+#else
+        QByteArray typeStr = re.cap(1).toLatin1().constData();
```
c42f: Qt4 has `toLatin1()` - maybe we just use that, or possibly better `toUtf8()`?  There shouldn't be any strange characters here, it's glsl source after all.

For everywhere we have the choice and must convert QString to bytes, I'd like to use utf8 for C strings in preference to the other possible encodings.  Again, not really relevant here I expect since this is glsl source.

src/shader.cpp
```diff
@@ -135,14 +143,17 @@ bool Shader::compileSourceCode(const QByteArray& src)
...
-    m_context(context),
     m_pointSize(5),
     m_exposure(1),
     m_contrast(1),
     m_selector(0)
-{ }
+{
+    m_vertexShader = NULL;
+    m_fragmentShader = NULL;
+    m_shaderProgram = NULL;
```
c42f: These should be unnecessary (actually I'm surprised they even work) - `unique_ptr` automatically initializes the held pointer to null.

src/shader.cpp
```diff
@@ -233,29 +244,37 @@ bool ShaderProgram::setShaderFromSourceFile(QString fileName)
...
-    std::unique_ptr<Shader> fragmentShader(new Shader(QGLShader::Vertex, m_context));
+    std::unique_ptr<Shader> vertexShader(new Shader(QGLShader::Fragment));
+    std::unique_ptr<Shader> fragmentShader(new Shader(QGLShader::Vertex));
     //tfm::printf("Shader source:\n###\n%s\n###\n", src.toStdString());
-    if(!vertexShader->compileSourceCode(src.toAscii()))
+#ifdef DISPLAZ_USE_QT4
+    QByteArray src_ba = src.toAscii();
+#else
+    QByteArray src_ba = src.toLatin1();
+#endif
```
c42f: Again, let's use `toUtf8`.  Unclear what we can usefully do if the source has odd characters.

src/shader.cpp
```diff
@@ -374,14 +401,6 @@ void ShaderProgram::setUniforms()
     }
 }
 
-
-void ShaderProgram::setContext(const QGLContext* context)
-{
-    m_context = context;
-    setShader(shaderSource());
-}
```
c42f: I admit I had some problems keeping the context around as `m_context`, and I was finding odd things happening.  I can see it's removed here - did you find anything enlightening about what was going wrong?

src/view3d.cpp
```diff
@@ -104,7 +117,15 @@ void View3D::geometryInserted(const QModelIndex& /*unused*/, int firstRow, int l
 {
     const GeometryCollection::GeometryVec& geoms = m_geometries->get();
     for (int i = firstRow; i <= lastRow; ++i)
-        geoms[i]->initializeGL();
+    {
+        if (m_boundingBoxShader->isValid())
+        {
+            geoms[i]->setShaderId("boundingbox", m_boundingBoxShader->shaderProgram().programId());
+            geoms[i]->setShaderId("meshface", m_meshFaceShader->shaderProgram().programId());
+            geoms[i]->setShaderId("meshedge", m_meshEdgeShader->shaderProgram().programId());
```
c42f: Looks like we'd be better off having a `ShaderCache` object or some such, to keep track of the set of compiled utility shaders.  That would separate some of the extra entangled logic out of `Geometry` / `View3D`, and make a central place to manage utility shaders.

src/view3d.cpp
```diff
@@ -185,31 +212,132 @@ void View3D::centerOnGeometry(const QModelIndex& index)
...
+        case GL_FRAMEBUFFER_UNSUPPORTED: status="UNSUPPORTED";      break;
+        default:                        status="INCOMPLETE";    break;
+    }
+
+    tfm::printfln("GL_FRAMEBUFFER_%s - %s:%i", status, file, line);
+}
+
+#define glFrameBufferStatus(TARGET) _glFrameBufferStatus(TARGET, __FILE__,__LINE__)
+
+#endif //GL_CHECK_RENDER
```
c42f: With an `#else` you can avoid any mention of `GL_CHECK_RENDER`, except where `glCheckError()` is defined:

```C++
#else
#define glCheckError()
#endif //GL_CHECK_RENDER

File: src/view3d.cpp [hunk 12]
```diff
@@ -185,31 +212,132 @@ void View3D::centerOnGeometry(const QModelIndex& index)
...
+
+    m_meshEdgeShader.reset(new ShaderProgram());
+    bool meshedge_shader_init = m_meshEdgeShader->setShaderFromSourceFile("shaders:meshedge.glsl");
+
     m_incrementalFramebuffer = allocIncrementalFramebuffer(width(), height());
+
+#ifdef GL_CHECK_RENDER
+    //glFrameBufferStatus(m_incrementalFramebuffer->handle());
+    glCheckError();
+#endif
```
c42f: With the above tweak, the `#ifdef` can be removed from here.

src/view3d.cpp
c42f: I think it's also worth moving this stuff into glutil.h since it seems like generally useful utility code.

src/view3d.cpp
```diff
@@ -271,23 +403,39 @@ void View3D::paintGL()
...
+            // boundingBoxShader.release();
+        }
     }
 
     // Draw meshes and lines
     if (!m_incrementalDraw)
     {
-        drawMeshes(transState, geoms);
-
+        this->drawMeshes(transState, geoms);
```
c42f: Purely as a matter of style consistency, could you remove the `this->` in various places?  It's not bad, just inconsistent with the rest of the code.

src/view3d.cpp
```diff
@@ -271,23 +403,39 @@ void View3D::paintGL()
...
         // Generic draw for any other geometry
         // (TODO: make all geometries use this interface, or something similar)
         // FIXME - Do generic quality scaling
         const double quality = 1;
         for (size_t i = 0; i < geoms.size(); ++i)
+        {
             geoms[i]->draw(transState, quality);
+        }
```
c42f: Another style nitpick: the codebase is fairly consistent about not using extra brackets around single line loops or if statements.  Could you kill these off?

src/corecontext.h
chrisidefix: Yes, this works ;) I added that `DISPLAZ_USE_QT4` flag to `CMake` and thought it should also work on OS X or do you want to remove Qt4 completely?

src/view3d.cpp
```diff
@@ -355,18 +518,18 @@ void View3D::drawMeshes(const TransformState& transState,
         meshFaceShader.setUniformValue("lightDir_eye", lightDir.x, lightDir.y, lightDir.z);
         for (size_t i = 0; i < geoms.size(); ++i)
             geoms[i]->drawFaces(meshFaceShader, transState);
-        meshFaceShader.release();
+        // meshFaceShader.release();
```
c42f: Is this unnecessary?  If so, let's remove it entirely rather than commenting it out.

src/view3d.cpp
```diff
@@ -448,92 +650,205 @@ void View3D::drawCursor(const TransformState& transStateIn, const V3d& cursorPos
     TransformState transState = transStateIn.translate(offset);
     V3d relCursor = cursorPos - offset;
 
-    // Cull if behind camera
-    if((relCursor * transState.modelViewMatrix).z > 0)
-        return;
+    // Cull if behind camera -> doesn't work for now
+    //if((relCursor * transState.projMatrix).z > 0)
+    //    return;
```
c42f: Any commentary here on what exactly was going wrong?  Did it work before and not now?  Was it just plainly broken before (entirely possible)?

src/mainwindow.cpp
```diff
@@ -479,6 +487,12 @@ void PointViewerMainWindow::addFiles()
     m_currFileDir.makeAbsolute();
 }
 
+void PointViewerMainWindow::finalSetup()
+{
+    openShaderFile("shaders:las_points.glsl");
```
c42f: Maybe let's put this in as the default inside `View3D`, rather than bothering with the extra complexity the `initializedGL` -> `finalSetup` signal slot connection.

src/corecontext.h
c42f: I want displaz to work with Qt4 on linux if possible.  The question is whether I should bother maintaining the Qt4 compatibility for OS X as well.  For windows, I'm inclined just to drop Qt4 completely since it's actually easier for developers to build against Qt5 on that platform (in particular, there's no official Qt4 binaries for windows).

Is that also the situation on OS X?  If so, it seems better just to say "it doesn't work with Qt4 on OS X - build with Qt5 instead - it's easier anyway".

c42f:
Ok, I finally had time to go through this, and I must say it's rather an epic effort - I think you went well beyond the call of duty here!

I did find a bunch of things which can be cleaned up.  Most should be relatively minor niggles, and some could potentially be put off (eg, creating a ShaderCache).

At a high level, I think the drawing code for small elements like axes and bounding boxes is clearly more convoluted now, which is rather disappointing but not really your fault.  I suspect the only way around this is some middleware to make the OpenGL core context nicer to use, or write our own utility code for the same thing.  I'm inclined to merge this sooner rather than later, and attempt to deal with the fallout on master...

As for hcloud, you can generate those from las with dvox, but it's pretty obscure and is experimental functionality in any case.  I'd be happy enough for you to ignore any brokenness there, and I'll have to clean it up later at some stage.

src/pointarray.cpp
chrisidefix: It was more of a suggestion, but I can see how this could make things more problematic on the CPU side. I think the only expected performance benefit would come from reducing the number of calls to `glDrawArrays`, which would also mean a reduced number of calls to `glBufferData`, but with larger buffer sizes, this might not be possible depending on how big that buffer is really going to get?

.travis.yml
```diff
@@ -29,12 +31,18 @@ before_script:
...
     # Build displaz
     - mkdir build
     - cd build
-    - cmake ..
+    - if [ "$TRAVIS_OS_NAME" == "linux" ]; then cmake -DDISPLAZ_USE_QT4=1 ..; fi
+    - if [ "$TRAVIS_OS_NAME" == "osx" ]; then cmake ..; fi
+    - make -j4
+    - make clean
+    - if [ "$TRAVIS_OS_NAME" == "linux" ]; then cmake -DCMAKE_BUILD_TYPE=Debug -DDISPLAZ_USE_QT4=1 ..; fi
+    - if [ "$TRAVIS_OS_NAME" == "osx" ]; then cmake -DCMAKE_BUILD_TYPE=Debug ..; fi
```
c42f: This seems like something that the travis build matrix should take care of?  Maybe we can we get one dimension of the build matrix to set an environment variable for CMAKE_BUILD_TYPE.

src/corecontext.h
chrisidefix: We can go with Qt5 on OS X. There is no need to keep Qt4 support around. There are official binaries for both versions, but Qt5 is mature enough on OS X to make it our requirement - it also fixes a number of bugs, which we don't have to deal with or maintain obscure workarounds for.

src/shader.cpp
chrisidefix: In my experience the actual source code isn't the problem when in comes to weird characters, but the comments can be. Sometimes weird characters can end up inside comments, but I guess `toUtf8()` would handle these cases for us as well.

src/shader.cpp
chrisidefix: I actually changed this, because the more recent `QOpenGLShader` doesn't allow you to set the context anymore and just uses the current one available, which also makes sense. It will be easier to switch to that class at some point in the future. I didn't see any particular problem with the previous implementation, though.

src/view3d.cpp
chrisidefix: I wanted to highlight that we should not "release" the shader, which is equivalent to calling `glUseProgram(0)`, at least that's what the Qt specs say. Should we keep that information in a better written comment or should we rather remove it everywhere?

src/shader.cpp
c42f: Sure, sounds reasonable.

src/view3d.cpp
chrisidefix: There was an issue with this check always returning `TRUE` and hence not rendering the cursor, because of some uninitialised variable somewhere. I think I fixed that along the way, but didn't go back try out this culling again. However, I don't see how the cursor could be behind the camera, though? Am I missing some interaction pattern here - the camera always centres around the cursor for me.

src/shader.cpp
c42f: Well, it's unclear whether shader compilers in typical drivers allow utf-8, but I suggest we assume they do until we see this causing problems in the wild.  The following link suggests utf-8 source encoding will be part of the future spec anyway:

https://www.khronos.org/bugzilla/show_bug.cgi?id=760



Ideally we'd just strip out any weird characters in comments, but `toAscii()` never really achieved that anyway.

src/view3d.cpp
c42f: I think remove it everywhere if that's the right thing to do.

My understanding is that `glUseProgram(0)` is harmless enough, and might be worthwhile in the sense that it prevents state from one part of the program from unexpectedly flowing on to another via the global OpenGL context: It's a way to mitigate the fact that the context is a global variable, to some extent.

Other than that, it's arguably an OpenGL API call that we don't need, so make things strictly less efficient.

Is there something else going on here that I don't understand?

src/view3d.cpp
c42f: True, it doesn't currently render behind the camera unless you use the undocumented ctrl+drag which lets you move it freely... but that's admittedly a bit useless as it stands and will be changed to something else at some stage.

src/view3d.cpp
```diff
@@ -78,7 +91,7 @@ View3D::View3D(GeometryCollection* geometries, QWidget *parent)
 
     m_incrementalFrameTimer = new QTimer(this);
     m_incrementalFrameTimer->setSingleShot(false);
-    connect(m_incrementalFrameTimer, SIGNAL(timeout()), this, SLOT(updateGL()));
+    // connect(m_incrementalFrameTimer, SIGNAL(timeout()), this, SLOT(updateGL()));
```
c42f: This should be put back as it was I suppose.

The incremental frame timer allows point clouds which are too large to be rendered in successive frames.  The visual effect is an unfortunate flickering, but it does give you a view of the whole dataset which can be quite important for technical work.

chrisidefix: Yeah, good catch. Done.

src/mainwindow.cpp
chrisidefix: Ok, done - have a look if you are ok with calling the parent window, otherwise more stuff has to be moved into View3D.

src/view3d.cpp
chrisidefix: This test is enabled again and it seems to work fine.

src/view3d.cpp
chrisidefix: Fair enough. Done.

src/view3d.cpp
chrisidefix: Sure. Done.

src/view3d.cpp
chrisidefix: Ok, I moved this into `glutil.h` and removed those `#ifdef` checks, but to actually enable glCheckError() calls, you still have to `#define GL_CHECK` for now.

chrisidefix:
Apart from your suggestions, I had to replace the Qt frame buffer setup with some native OpenGL calls, since the Qt version just kept messing up the render window. I also added a `setFocus()` for the render window to refresh it on start-up.

Would be good if you could check how this looks under linux / windows with Qt4 / Qt5.

src/view3d.cpp
c42f: Makes perfect sense, thanks.

src/view3d.cpp
```diff
@@ -185,61 +209,155 @@ void View3D::centerOnGeometry(const QModelIndex& index)
...
+
+    m_meshEdgeShader.reset(new ShaderProgram());
     m_meshEdgeShader->setShaderFromSourceFile("shaders:meshedge.glsl");
-    m_incrementalFramebuffer = allocIncrementalFramebuffer(width(), height());
+
+#ifdef DISPLAZ_USE_QT4
+    double dPR = 1.0;
+#else
+    double dPR = devicePixelRatio();
+#endif
```
c42f: This `#ifdef` stuff is repeated in several places, which isn't good for maintainability.  Can you refactor it into a utility function?

src/view3d.cpp
```diff
@@ -311,37 +461,42 @@ void View3D::paintGL()
 //        s[barSize] = '|';
 //    tfm::printfln("%12f %4d %s", quality, frameTime, s);
 
-    m_incrementalFramebuffer->release();
-    QGLFramebufferObject::blitFramebuffer(0, QRect(0,0,width(),height()),
-                                          m_incrementalFramebuffer.get(),
-                                          QRect(0,0,width(),height()),
-                                          GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+    // TODO: this should really render a texture onto a quad and not use glBlitFramebuffer
```
c42f: Why?  In principle `glBlitFramebuffer` should be strictly faster, the intention is to trigger a simple `memcpy()` of device memory.  Is it slower in practice because the driver writers don't optimize this path?

src/view3d.cpp
```diff
@@ -250,11 +368,30 @@ void View3D::paintGL()
...
+#ifdef DISPLAZ_USE_QT4
+    double dPR = 1.0;
+#else
+    double dPR = devicePixelRatio();
+#endif
+    int w = width() * dPR;
+    int h = height() * dPR;
+
+    // detecting a change in the device pixel ratio, since only the new QWindow (Qt5) would
+    // provide a signal for screen changes and this is the easiest solution
```
c42f: Eek, sounds nasty.  Good job hunting it down.

src/view3d.cpp
```diff
@@ -185,61 +209,155 @@ void View3D::centerOnGeometry(const QModelIndex& index)
...
+
+    glGenFramebuffers(1, &fb);
+    glBindFramebuffer(GL_DRAW_FRAMEBUFFER, fb);
+
     // TODO:
     // * Should we use multisampling 1 to avoid binding to a texture?
-    const QGLFormat fmt = context()->format();
-    QGLFramebufferObjectFormat fboFmt;
-    fboFmt.setAttachment(QGLFramebufferObject::Depth);
+    // * but why would you want to avoid binding to a texture? It should be more efficient than copying the fb
```
c42f: Let's have the conversation about this in the pull request, rather than the code comments :-)

c42f:
Quick test on linux for the new framebuffer changes - Qt5 version works fine. Qt4 goes into an infinite loop and tries to allocate all the memory on the machine!  Don't have time to dig into why right now, but for reference version 8802383 worked ok.

c42f:
Another minor thing to fix - there's various compiler warnings now generated by gcc - see the travis build log.  Normally it builds warning free for me (not quite sure about on travis), so it would be worth fixing most of those (`-Wint-to-pointer-cast` `-Wreorder` `-Wcomment` `-Wunused-variable` at least)

src/view3d.cpp
chrisidefix: Technically, it could probably be faster, but the tests I have read in the past claim it is slower (not a huge difference, though, maybe 20 microseconds). We should probably test this on current drivers with targeted hardware if we feel there is need for this to change.

c42f: Right.  There's an extra complexity here - the depth should be copied as well as the colour.  How do you deal with that using a texture, without resorting to `gl_FragDepth`?

chrisidefix:
I actually always got quite a few warnings, most of them being: `../dist/include/OpenEXR/ImathMatrix.h:2534:5: warning: 'register' storage class specifier is deprecated [-Wdeprecated-register]`

Those warning should be gone now and I think the infinite loop in Qt4 should also be fixed.

src/view3d.cpp
chrisidefix: Yes, done. This situation is not ideal in any case, and another good reason to switch to Qt5 sooner than later.

src/view3d.cpp
chrisidefix: Yeah, I would have preferred to use the `screenChanged()` signal, but we would have to move away from `QMainWindow`, which seems like a bit of work.

src/view3d.cpp
chrisidefix: You can attach color textures and a depth texture (plus a stencil texture) to the frame buffer:
`glFramebufferTexture(GL_DRAW_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, colorTex, 0);`
`glFramebufferTexture(GL_DRAW_FRAMEBUFFER,  GL_DEPTH_ATTACHMENT, depthTex, 0);`
The depth texture would be setup with a single `GL_DEPTH_COMPONENT` (probably with `GL_FLOAT` format). This can be very useful when you want to do screen-space ambient occlusion, for example.

c42f: Right, so that lets you render the point cloud depth incrementally to a depth texture.  However, what we want to do here is copy that depth texture into the driver controlled backbuffer, followed by rendering extra stuff like bounding boxes to the normal backbuffer, with depth testing turned on.  I still don't see how to do that in a natural way.

src/view3d.cpp
c42f: Sure, makes sense.

src/glutil.cpp
```diff
@@ -32,61 +47,80 @@ void TransformState::setUniforms(GLuint prog) const
...
-                     const Imath::Box3d& bbox,
-                     const Imath::C3f& col)
+void drawBox(const TransformState& transState,
+             const Imath::Box3d& bbox,
+             const Imath::C3f& col,
+             const GLuint& shaderProgram)
 {
     // Transform to box min for stability with large offsets
+    // This function is allowing you to render any box, not necessarily having anything to do
+    // with the bounding box (e.g. hcloudview is using this in drawBounds)
```
c42f: Sure, no problems renaming it for clarity.

A quibble about this kind of comment - someone reading it later won't have the context of knowing the function was originally used to draw a bounding box (or was indeed called `drawBoundingBox()`), so talking about bounding boxes here will probably just confuse them.  I try to avoid comments which can only be understood in the context of the history of the codebase.

src/glutil.cpp
```diff
@@ -32,61 +47,80 @@ void TransformState::setUniforms(GLuint prog) const
...
+    tfm::printfln("drawBox :: has not been implemented, yet.");
+}
 
 void drawBoundingBox(const TransformState& transState,
-                     const Imath::Box3f& bbox,
-                     const Imath::C3f& col)
+                     const GLuint& bboxVertexArray,
+                     const Imath::V3f& offset,
+                     const Imath::C3f& col,
+                     const GLuint& shaderProgram)
```
c42f: This seems to have become something more like `draw12SmoothLines()` at this stage ;-)  Looks like I'll need to refactor the drawing code into a bunch of utilities... using core context only really is a different ballgame.

Thanks for all the work here, I've learned quite a bit about modern OpenGL while looking through it.

src/view3d.cpp
```diff
@@ -185,61 +209,156 @@ void View3D::centerOnGeometry(const QModelIndex& index)
...
+        {
+            // TODO: build a shader manager for this
+            geoms[i]->setShaderId("boundingbox", m_boundingBoxShader->shaderProgram().programId());
+            geoms[i]->setShaderId("meshface", m_meshFaceShader->shaderProgram().programId());
+            geoms[i]->setShaderId("meshedge", m_meshEdgeShader->shaderProgram().programId());
+            geoms[i]->initializeGL();
+        }
+    }
+
+    PointViewerMainWindow * pv_parent = dynamic_cast<PointViewerMainWindow *>(parentWidget());
```
c42f: Oh dear, so this is why you needed initializedGL()... so that the shader editor would also get the correct shader without something horrible like this `dynamic_cast`.  Sorry about that... can be cleaned up later.  Looks like I should connect up the text editor's `QTextDocument` to the shader or some such thing - some kind of more direct connection rather than proxying through the `PointViewMainWindow`...

src/view3d.cpp
```diff
@@ -185,61 +209,156 @@ void View3D::centerOnGeometry(const QModelIndex& index)
...
+    double dPR = getDevicePixelRatio();
+
+    m_camera.setViewport(QRect(0,0,double(w)/dPR,double(h)/dPR));
+
     m_incrementalFramebuffer = allocIncrementalFramebuffer(w,h);
+
+    glCheckError();
+
+    // without this explicit call, we get very strange artifacts during resizing
+    //updateGL();
```
c42f: Is this still true, or did you sort it out?

c42f:
This seems really close now, I think there's just a final bit of cleanup to do: scrolling through I can see a bunch of pieces of code which are commented out, but without an associated TODO or comment explaining why I might need them in the future.  Could you make a final pass and either remove the commented pieces, or add a quick note/TODO for the ones which don't have an obvious use.

src/view3d.cpp
chrisidefix: This was true, while using the Qt Framebuffer and I didn't see this problem anymore after switching to native OpenGL calls.

src/view3d.cpp
chrisidefix: Yes, this certainly isn't the prettiest solution, but it's a bit of work to restructure all of this.

c42f: Yes, definitely work for later.

chrisidefix:
Alright, I didn't adjust the travis setup, because it should be switched to using the newer `sudo: false` container-based infrastructure anyway, which is supposed to build faster: http://docs.travis-ci.com/user/workers/container-based-infrastructure/

Unless you find some important problem, feel free to test and merge at your convenience ;)

c42f:
Yes, I think this is good to go in principle, just going to do some additional testing.

nigels-com:
There is a little fix needed for the Windows build, which I'm try out just now.

nigels-com:
https://github.com/geometree/displaz/pull/2

nigels-com:
I see a crash on Win7 for VC12, Qt 4.8.6, x86 release mode build.

c42f:
Does it work with Qt5?  I'd be inclined to say the windows build is Qt5 only.

nigels-com:
Oh, right.  You did say that!  Qt5 time, I guess.

nigels-com:
No clues from valgrind on Linux, with Qt4 or Qt5, unfortunately.  I'll try Qt5 on Windows next.

nigels-com:
Hmmm....  Doesn't look healthy...

```
==8474== Thread 6 QThread:
==8474== Invalid read of size 4
==8474==    at 0x55014D: LASindex::read(char const*) (in /usr/local/bin/displaz-gui)
==8474==    by 0x5214D0: LASreadOpener::open(char const*) (in /usr/local/bin/displaz-gui)
==8474==    by 0x4BD9BB: PointArray::loadLas(QString, unsigned long, std::vector<GeomField, std::allocator<GeomField> >&, Imath::Vec3<double>&, unsigned long&, unsigned long&, Imath::Box<Imath::Vec3<double> >&, Imath::Vec3<double>&) (in /usr/local/bin/displaz-gui)
==8474==    by 0x4C1609: PointArray::loadFile(QString, unsigned long) (in /usr/local/bin/displaz-gui)
==8474==    by 0x493FF9: FileLoader::qt_static_metacall(QObject*, QMetaObject::Call, int, void**) (in /usr/local/bin/displaz-gui)
==8474==    by 0x4FCCC1D: QObject::event(QEvent*) (in /usr/lib/x86_64-linux-gnu/libQtCore.so.4.8.6)
==8474==    by 0x54E5E2B: QApplicationPrivate::notify_helper(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQtGui.so.4.8.6)
==8474==    by 0x54EC49F: QApplication::notify(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQtGui.so.4.8.6)
==8474==    by 0x4FB44DC: QCoreApplication::notifyInternal(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQtCore.so.4.8.6)
==8474==    by 0x4FB7B3C: QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) (in /usr/lib/x86_64-linux-gnu/libQtCore.so.4.8.6)
==8474==    by 0x4FE1F82: ??? (in /usr/lib/x86_64-linux-gnu/libQtCore.so.4.8.6)
==8474==    by 0x7BC9E03: g_main_context_dispatch (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4002.0)
==8474==  Address 0x1df26a3c is 44 bytes inside a block of size 47 alloc'd
==8474==    at 0x4C2AB80: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==8474==    by 0x7209839: strdup (strdup.c:42)
==8474==    by 0x550130: LASindex::read(char const*) (in /usr/local/bin/displaz-gui)
==8474==    by 0x5214D0: LASreadOpener::open(char const*) (in /usr/local/bin/displaz-gui)
==8474==    by 0x4BD9BB: PointArray::loadLas(QString, unsigned long, std::vector<GeomField, std::allocator<GeomField> >&, Imath::Vec3<double>&, unsigned long&, unsigned long&, Imath::Box<Imath::Vec3<double> >&, Imath::Vec3<double>&) (in /usr/local/bin/displaz-gui)
==8474==    by 0x4C1609: PointArray::loadFile(QString, unsigned long) (in /usr/local/bin/displaz-gui)
==8474==    by 0x493FF9: FileLoader::qt_static_metacall(QObject*, QMetaObject::Call, int, void**) (in /usr/local/bin/displaz-gui)
==8474==    by 0x4FCCC1D: QObject::event(QEvent*) (in /usr/lib/x86_64-linux-gnu/libQtCore.so.4.8.6)
==8474==    by 0x54E5E2B: QApplicationPrivate::notify_helper(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQtGui.so.4.8.6)
==8474==    by 0x54EC49F: QApplication::notify(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQtGui.so.4.8.6)
==8474==    by 0x4FB44DC: QCoreApplication::notifyInternal(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQtCore.so.4.8.6)
==8474==    by 0x4FB7B3C: QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) (in /usr/lib/x86_64-linux-gnu/libQtCore.so.4.8.6)
```

nigels-com:
The PACKAGE step for Windows/Qt5 doesn't seem to work yet.  I'll take a closer look tomorrow.  Meanwhile I put in another merge request for Windows/Qt/cmake.

```
2>------ Build started: Project: PACKAGE, Configuration: Release x64 ------
2>  CPack: Create package using NSIS
2>  CPack: Install projects
2>  CPack: - Install project: displaz
2>EXEC : warning : target 'Qt5OpenGL.dll' is not absolute...
2>EXEC : warning : target 'Qt5OpenGL.dll' does not exist...
2>EXEC : warning : target 'Qt5Network.dll' is not absolute...
2>EXEC : warning : target 'Qt5Network.dll' does not exist...
2>EXEC : warning : target 'Qt5Widgets.dll' is not absolute...
2>EXEC : warning : target 'Qt5Widgets.dll' does not exist...
2>EXEC : warning : target 'Qt5Gui.dll' is not absolute...
2>EXEC : warning : target 'Qt5Gui.dll' does not exist...
2>EXEC : warning : target 'Qt5Core.dll' is not absolute...
2>EXEC : warning : target 'Qt5Core.dll' does not exist...
2>EXEC : warning : target 'Qt5Network.dll' is not absolute...
2>EXEC : warning : target 'Qt5Network.dll' does not exist...
2>EXEC : warning : target 'Qt5Core.dll' is not absolute...
2>EXEC : warning : target 'Qt5Core.dll' does not exist...
2>  Error copying file "Qt5Core.dll" to "C:/Users/nigels/dev/displaz/build/_CPack_Packages/win64/NSIS/displaz-0.3.1-166-g1ec5cdb-win64/bin/Qt5Core.dll".
2>  
```

c42f:
@nigels-com - what's the current status on windows, does this work with either Qt4 or Qt5?

nigels-com:
In a nutshell, Qt4 builds and packages, but crashes at runtime.  Qt5 builds and runs OK, but can't package.  I was going to play with stack size in the Qt4 build, are there any big arrays on the stack in recent changes?

c42f:
Right thanks.  I'm inclined to try bashing a bit more at cmake to get Qt5 packaging properly.  Will try that as soon as I get time.  Maybe related to https://public.kitware.com/Bug/view.php?id=15400 , though they're using `BundleUtilities` rather than `InstallRequiredSystemLibraries`

c42f:
I figured out the problem with packaging on windows - there's no such variable as `Qt5_Core_LIBRARY_DIR` which was used in the fixup_bundle step.  Pull request coming your way.

Unfortunately I still get a hard crash immediately on launching displaz on windows.  This may be because I only have a VM for testing at the moment though.

c42f:
@nigels-com if you have time, could you build & do a sanity check on windows with qt5?  I can build and package locally for windows but displaz crashes in my VM.

I really should set up a non virtualized test box.  I did a bit of reading and it seems like the only obviously reliable option for OpenGL 3.3 at this stage is to set up a dual boot.

nigels-com:
Yes, will do.

c42f:
Cheers, thanks.

Separate note: I also noticed that the icon no longer seems to come up properly when installed on windows.  Ah, the joys of cross platform dev :-(

nigels-com:
The good news is that it builds, packages and installs for Qt5 on Windows 7.  Also, it runs without crashing (so far!), which is promising indeed.  However, I'm still seeing a crummy console pop-up, which might have something to do with the icon issue.  Is it the displaz/displaz-gui dichotomy?

c42f:
Yup, it's the displaz/displaz-gui split which will have caused it the black window - it's a bug unrelated to this PR (see #84).  I think you're right and the icon trouble is very likely related to #84 as well.

c42f:
Well since the win32 smoke test now passes, I'm mightily tempted to just merge this - it's holding up further work at this stage.

PR #80, opened 2015-10-17
Fixup for debug-mode build

nigels-com:


c42f:
Ah thanks.  Looks like I forgot to refactor the asserts at some stage and I don't build it every time in debug mode (could be something to set up on travis?)

I'll commit an alternative version which will just delete this line - the invariant (with a different name for the variable) is now enforced in the constructor instead.

PR #81, opened 2015-10-19
GLEW: Update to upstream release 1.13.0

nigels-com:


c42f:
I'm happy to merge this.  For my own education, are there any parts of this which obviously effect displaz itself in a move to the 3.2 core profile?

nigels-com:
There is nothing in 1.13.0 that ought to impact GLEW for either compatibility or core profiles.  But the experimental core profile branch is relative to this, so no need to have the additional churn in a displaz dev branch.

See also: http://glew.sourceforge.net/log.html

c42f:
Ah, that makes sense, thanks.

PR #93, opened 2016-02-24
New Remote Commands

landersson:
Hi Chris, here's the changes I've made to add a couple of new commands that I needed, plus a fix to an assert statement that seems to trigger erroneously when loading some of my ply files in debug mode.

### DIFF
--- a/src/geometry.cpp
+++ b/src/geometry.cpp
@@ -51,7 +51,9 @@ Geometry::Geometry()
 
 Geometry::~Geometry()
 {
-    destroyBuffers();
+    // Line below commented out as a temporary fix to prevent segfaults 
+    // when removing datasets. Probably leaking memory now instead though.
+    //destroyBuffers();
 }
 
 std::shared_ptr<Geometry> Geometry::create(QString fileName)
--- a/src/geometry.cpp
+++ b/src/geometry.cpp
@@ -51,9 +51,7 @@ Geometry::Geometry()
 
 Geometry::~Geometry()
 {
-    // Line below commented out as a temporary fix to prevent segfaults 
-    // when removing datasets. Probably leaking memory now instead though.
-    //destroyBuffers();
+    destroyBuffers();
 }
 
 std::shared_ptr<Geometry> Geometry::create(QString fileName)
--- a/src/geometrycollection.cpp
+++ b/src/geometrycollection.cpp
@@ -23,11 +23,11 @@ void GeometryCollection::clear()
     }
 }
 
-int GeometryCollection::findMatchingRow(const QRegExp & filename_regex)
+int GeometryCollection::findMatchingRow(const QRegExp & filenameRegex)
 {
     for (unsigned row = 0; row < m_geometries.size(); row++)
     {
-        if (filename_regex.exactMatch(m_geometries[row]->fileName()))
+        if (filenameRegex.exactMatch(m_geometries[row]->fileName()))
         {
             return row;
         }
@@ -35,11 +35,11 @@ int GeometryCollection::findMatchingRow(const QRegExp & filename_regex)
     return -1;
 }
 
-void GeometryCollection::unloadFiles(const QRegExp & filename_regex)
+void GeometryCollection::unloadFiles(const QRegExp & filenameRegex)
 {
     while (true)
     {
-        int row = findMatchingRow(filename_regex);
+        int row = findMatchingRow(filenameRegex);
 
         if (row == -1)
             break;
--- a/src/geometrycollection.cpp
+++ b/src/geometrycollection.cpp
@@ -6,6 +6,8 @@
 #include "tinyformat.h"
 #include "fileloader.h"
 
+#include <QRegExp>
+
 GeometryCollection::GeometryCollection(QObject* parent)
     : QAbstractListModel(parent)
 { }
@@ -21,11 +23,11 @@ void GeometryCollection::clear()
     }
 }
 
-int GeometryCollection::findMatchingRow(const std::string & filename_substr)
+int GeometryCollection::findMatchingRow(const QRegExp & filename_regex)
 {
     for (unsigned row = 0; row < m_geometries.size(); row++)
     {
-        if (std::string::npos != m_geometries[row]->fileName().toStdString().find(filename_substr))
+        if (filename_regex.exactMatch(m_geometries[row]->fileName()))
         {
             return row;
         }
@@ -33,11 +35,11 @@ int GeometryCollection::findMatchingRow(const std::string & filename_substr)
     return -1;
 }
 
-void GeometryCollection::unloadFiles(const std::string & filename_substr)
+void GeometryCollection::unloadFiles(const QRegExp & filename_regex)
 {
     while (true)
     {
-        int row = findMatchingRow(filename_substr);
+        int row = findMatchingRow(filename_regex);
 
         if (row == -1)
             break;
--- a/src/geometrycollection.cpp
+++ b/src/geometrycollection.cpp
@@ -21,6 +21,34 @@ void GeometryCollection::clear()
     }
 }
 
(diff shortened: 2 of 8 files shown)


src/geometrycollection.cpp
```diff
@@ -21,6 +21,34 @@ void GeometryCollection::clear()
     }
 }
 
+int GeometryCollection::findMatchingRow(const std::string & filename_substr)
+{
+    for (unsigned row = 0; row < m_geometries.size(); row++)
+    {
+		if (std::string::npos != m_geometries[row]->fileName().toStdString().find(filename_substr))
```
nigels-com: Looks like indentation needs cleaning up.

src/main.cpp
```diff
@@ -85,9 +87,11 @@ int main(int argc, char *argv[])
         "-noserver",     &noServer,      "Don't attempt to open files in existing window",
         "-server %s",    &serverName,    "Name of displaz instance to message on startup",
         "-shader %s",    &shaderName,    "Name of shader file to load on startup",
+        "-viewposition %F %F %F", &posX, &posY, &posZ, "Set absolute view position [X, Y, Z]",
```
nigels-com: That reminds me, I'm getting hungrier for ortho projection mode.. :-)

src/geometry.cpp
```diff
@@ -51,7 +51,9 @@ Geometry::Geometry()
 
 Geometry::~Geometry()
 {
-    destroyBuffers();
+    // Line below commented out as a temporary fix to prevent segfaults 
+    // when removing datasets. Probably leaking memory now instead though.
+    //destroyBuffers();
```
c42f: Could you put this back since it's a separate issue, and this obviously leaking isn't a viable solution (by all means, keep the hack in your local copy until we can get to the root cause).

src/geometrycollection.cpp
```diff
@@ -21,6 +21,34 @@ void GeometryCollection::clear()
...
+        {
+            printf("MATCHED '%s'\n", m_geometries[row]->fileName().toStdString().c_str());
```
c42f: Use `tfm::printf` and you won't need the `c_str()` ;-)

Regardless of that, let's remove the printf debug logging before merging this.

src/main.cpp
```diff
@@ -215,6 +219,17 @@ int main(int argc, char *argv[])
...
+    if (!unloadFiles.empty())
+    {
+        channel->sendMessage(QByteArray("UNLOAD_FILES\n") + unloadFiles.c_str());
+    }
+    if (posX != -DBL_MAX)
+    {
+        channel->sendMessage("SET_VIEW_POSITION\n" +
+                             QByteArray().setNum(posX, 'g', 32) + "\n" +
+                             QByteArray().setNum(posY, 'g', 32) + "\n" +
+                             QByteArray().setNum(posZ, 'g', 32));
```
c42f: I think these would make most sense as `setNum(posX, 'e', 17)` which is guaranteed to reproduce the incoming floating point number exactly, barring special values like NaN, Inf.

c42f:
I've had a quick look through, thanks a lot for doing this and making the effort to create a PR.

I'm definitely convinced by the need for a `-viewposition` option and I like what you've got here.

The best semantics of the `-unload` option are a bit less obvious, and in particular what kind of pattern matching of the name should occur (maybe `QRegExp::WildcardUnix` would be better?)  We should also try to make this stuff work with non-ascii character sets where possible, which I prefer to mean using UTF8 over the socket and QString elsewhere in the UI part of the code.  (Having said that, the command line parsing probably isn't unicode safe on windows... I'm guilty as charged on that I'm afraid.)

c42f:
By the way, there's probably a race condition with `-viewposition` since SET_VIEW_POSITION will be executed immediately before any files are loaded, which happens in the background in a separate thread... and the first file loaded will then set the cursor position to the center of its bounding box.

I'm more or less happy to leave that to fix later, but if you have the inclination to tackle it, I think there should be some kind of flag which tracks whether `-viewposition` was used prior to the first file loading.

landersson:
Thanks for the review and comments, and sorry about wasting your time on looking at debug printf's! I had cleaned it up, but as it turns out, I accidentally cleaned up the code in my original clone of your repo rather than in my fork.

I actually used c++11 regex's in my original code for the "-unload" command, but then discovered that the regex class is not fully supported by g++4.8 and then opted for a simple substring hack that would solve my immediate problems. Since we already depend on QT, using their regex class is obviously the way to go and I've added support for that.

I'm not quite sure what the best way of solving the -viewposition race condition is, but I've added something that probably works better than nothing. I'll let the code speak for itself, but the end result is that after specifying an explicit position using the SET_VIEW_POSITION command, we wont automatically center on geometry again after that, except when manually double clicking on a dataset in the list. I suppose that is roughly the behaviour we want. 

I guess there is still a race condition in there though that will hit if files load very fast and trigger the geometryLoaded signal before the SET_VIEW_POSITION command is executed. I haven't tried to solve that issue.

src/mainwindow.cpp
```diff
@@ -378,6 +378,37 @@ void PointViewerMainWindow::handleMessage(QByteArray message)
     {
         m_geometries->clear();
     }
+    else if (commandTokens[0] == "UNLOAD_FILES")
+    {
+        QString regex_str = commandTokens[1];
+        QRegExp regex(regex_str, Qt::CaseSensitive, QRegExp::WildcardUnix);
+        if (!regex.isValid())
+        {
+            tfm::format(std::cerr, "Invalid regular expression '%s': %s\n",
```
c42f: This should probably go to `g_logger` so the user has a better chance of seeing it.

Ideally we'd have a stronger IPC format which would allow the sender of the message to get an ACK with success or failure to report on the command line rather than in the GUI... but that's for another time.

src/mainwindow.cpp
```diff
@@ -378,6 +378,37 @@ void PointViewerMainWindow::handleMessage(QByteArray message)
...
+            return;
+        }
+        m_geometries->unloadFiles(regex);
+    }
+    else if (commandTokens[0] == "SET_VIEW_POSITION")
+    {
+        if (commandTokens.size()-1 != 3)
+        {
+            tfm::format(std::cerr, "Expected three coordinates, got %d\n",
+                        commandTokens.size()-1);
```
c42f: I think this is fine going to cerr btw (in contrast to the regex above) - if this error ever happens it's most probably an internal bug in displaz rather than user error.

c42f:
This looks good, thanks for tackling the race condition.  The best behaviour after setting the view position is unclear to me: in some circumstances the user may want displaz to unset the m_explicitCursorPos after loading the next file so that auto centering is re-enabled.  In other circumstances I can see this being equally frustrating.  I guess we'll try out what you've implemented and see :)

I was about to merge this when I found one more niggle: the matching is done on filename.  However, the file name is not what users see in the displaz interface - instead they see the dataset name.  This can be any user defined label using the "-dataname" option.  I think -unload should potentially match against these user defined names instead (these names default to the file name), so I've made this tweak which I think will also work for you.  Please let me know if that doesn't work though!

In addition to the file name and the dataset label I think there's scope for a fully unique dataset ID, but we'll do that later if necessary.

(In hindsight, why didn't I call the `-dataname` option `-label`? ugh... I think I'm going to change this right away before it gets any more embedded.)

landersson:
Great, thanks for merging this. My rationale for using fileName() was that the unload option would probably be used from a script in most cases, where filenames may be a more natural reference, but since name() defaults to fileName(), that works for me.

c42f:
Right.  The issue with file names is that displaz can have the same file open multiple times, which is fairly natural when running the same program multiple times to produce different outputs.  In this case the file name is far from unique, and a label is required.  OTOH, the label is also not guaranteed to be unique since it's user defined.  As I mentioned above, there's probably scope for something like a uuid to programmatically address a given dataset in memory, but I'll add that later as the use case becomes clearer.

PR #94, opened 2016-02-26
Send initial point shader on displaz command line

landersson:
I just ran into the following problem: I wanted to adjust initial point radius, exposure etc so I copied the original las_points.glsl, made some changes, and specified my new glsl file on the command line using the -shader option. However, my changes to these parameters didn't have any effect, because the default point shader was loaded at start up time in view3d.cpp, and because of lines 346-347 in shader.cpp, the parameters were not updated when displaz-gui received the OPEN_SHADER command to open my new shader file. 

I don't quite understand the reason for lines 346-347 in shader.cpp and my initial quick fix was to simply comment them out. Then, looking a bit closer at the OPEN_SHADER command, I realized that it's only executed when starting a new instance of displaz-gui anyway. Maybe options that are only valid at gui start up should be passed to displaz-gui as command line parameters rather than commands? In any case, feel free to merge this pull request if it makes sense to you.

### DIFF
--- a/src/guimain.cpp
+++ b/src/guimain.cpp
@@ -43,12 +43,14 @@ int main(int argc, char* argv[])
     std::string lockName;
     std::string lockId;
     std::string socketName;
+    std::string shaderName;
 
     ArgParse::ArgParse ap;
     ap.options(
         "displaz-gui - don't use this directly, use the displaz commandline helper instead)",
         "-instancelock %s %s", &lockName, &lockId, "Single instance lock name and ID to reacquire",
         "-socketname %s", &socketName,             "Local socket name for IPC",
+        "-shader %s",    &shaderName,              "Name of shader file to load on startup",
         NULL
     );
 
@@ -79,7 +81,7 @@ int main(int argc, char* argv[])
     //f.setSampleBuffers(true);
     QGLFormat::setDefaultFormat(f);
 
-    PointViewerMainWindow window(f);
+    PointViewerMainWindow window(f, QString::fromStdString(shaderName));
     InterProcessLock instanceLock(lockName);
     if (!lockId.empty())
         instanceLock.inherit(lockId);
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -163,7 +163,9 @@ int main(int argc, char *argv[])
         QStringList args;
         args << "-instancelock" << QString::fromStdString(lockName)
                                 << QString::fromStdString(instanceLock.makeLockId())
-             << "-socketname"   << socketName;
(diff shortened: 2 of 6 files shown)


c42f:
Thanks!  Just a quick note that I'm going on holidays tomorrow, so if I don't manage to look at this for a week and a half don't take this as a sign of disinterest :)

c42f:
> I don't quite understand the reason for lines 346-347 in shader.cpp and my initial quick fix was to simply comment them out

The reason for these lines is that the user can interactively recompile the shader from displaz using the shader editor (shift+Return).  When they do this, the shader parameter UI needs to be rebuilt because the UI controls are defined by comments in the shader itself.  However, if the values of the shader controls in the UI were reset to the defaults, this is really annoying when interactively recompiling to incrementally develop the shader.  For this reason, the previous version of the values are kept.

Now, all of the above was designed early in development when there was only one shader available and no scripting interface.  Nowadays the assumption that you *always* want to keep the previous shader parameters is broken, as you found out.

Keeping all of that in mind, what you've got here is not my preferred fix - I've intentionally made all parameters pass through the IPC channel for consistency and this is a *lot* cleaner than half a year ago or so when a subset of the parameters went via the command line.

Here's what I'd do for a relatively easy fix: I'd change `ShaderProgram::setShader()` so that it takes an optional bool parameter specifying whether to keep previous shader parameters.  Set this to false by default, but set it to true when calling from `PointViewerMainWindow::compileShaderFile()`, which is triggered by the interactive recompile keyboard shortcut.  Might also be worth surveying the other uses of `setShader()` to determine whether each should keep the existing defaults or not.

As you can see by the TODOs, the whole shader setting thing ought to be more coherent.  One thing I'll eventually do is the ability to watch shaders for changes on the filesystem so that a more reasonable external text editor can be used instead of the builtin `QPlainTextEdit` which is horrible to use as a code editor.

landersson:
Ok, no problem. I actually started out implementing a fix pretty much exactly along the lines of your suggestion above, before I decided avoiding to load the wrong shader from the start might be better. Anyway, I just went on to two weeks of paternity leave and will also be a bit (very) preoccupied the next few weeks, but I'll send a PR based on your suggestion when I get around to it.

c42f:
Thanks, I totally understand that type of preoccupation, it's pretty fresh in my mind :)

Let's close this PR for now, and you can open a new one since the approach will be largely different.  If you can think of a better way to solve your problem then by all means go for it.  I'd definitely like the GUI process to get its commands via the same code path regardless of whether it's just been exec'd or is a persistent window hanging around.

Issue #74, opened 2015-08-22
Source code restructuring


c42f:
The code has grown to a size where it's worth thinking about structuring things a bit more cleanly for clarity.  Any larger branches should be merged first to avoid unnecessary pain.

Grab bag of tasks:

- [ ] Consider renaming some of the older classes, for example PointArray
- [ ] Rename files after the classes contained therein where it makes sense, including capitalization.
- [ ] Consider whether having some subdirectories makes sense.  Potential categories include:
  * general GUI stuff
  * 3D rendering
  * file input
  * the dvox tool and associated hcloud utils
  * system utils (IPC, logging, etc)

PR #96, opened 2016-03-09
Reorganise GUI and OpenGL rendering code into subdirectories

nigels-com:
Regarding https://github.com/c42f/displaz/issues/74

No functional change, just a shuffle.


c42f:
Looks reasonable.  I'm not sure where View3d belongs, since it's partly a Qt widget and partly a manager for OpenGL handles (probably a sign it needs refactoring?)  I guess the general guideline is that `gui` contains pure Qt GUI code, while `render` contain mostly OpenGL API calls + a qt utility glue as required?

To save on renaming churn, can I bother you to name the new files with the name of the main class inside (including case)?  I've been slowly moving in that direction when adding new classes, but haven't done a mass rename yet.

nigels-com:
Yes, I was aiming for "pure Qt" in GUI, versus "rendering related including some Qt stuff" in render.  I makes sense that there are places combining those worlds.  As for renaming, sure thing, will do.

nigels-com:
Done.

c42f:
Merged, thanks!

PR #98, opened 2016-03-21
Screenshot Command

landersson:
First take on a remote command for saving a screenshot of the current view.


c42f:
Thanks a lot, the general structure of this looks great.

I think there's a bit of work left to do to make sure that taking screenshots doesn't effect the GUI and OpenGL state, in particular, the camera handling.  Will have a closer look through the code.

Also I'm not sure about controlling what gets rendered into the screenshot.  In principle you might want to render bounding boxes, meshes, cursor, etc.  Probably the easiest thing is to take this state from the GUI so that the rendered screenshot resembles the GUI as much as possible (see also the camera location).

src/render/View3D.cpp
```diff
@@ -498,6 +514,42 @@ void View3D::paintGL()
...
+    if (height < 0)
+        height = this->height();
+
+    // Note: the reason for the following two lines and the subsequent downsampling of the
+    //       resulting image is to make the radius of rendered points consistent between the
+    //       gui and screenshot on displas with dPR != 1.
+    width *= dPR;
+    height *= dPR;
+
+    QOpenGLFramebufferObject framebuffer(QSize(width, height), QOpenGLFramebufferObject::Depth);
```
c42f: This class is Qt5 only, so we'd be breaking with Qt4 compatibility by using this.  I guess let's get it all working, and then I'll need to decide.

(TBH, I'm kinda keen to give up on Qt4 because the number of different OpenGL and OS versions is already enough to be a real drag on limited development resources...)

src/render/View3D.cpp
```diff
@@ -498,6 +514,42 @@ void View3D::paintGL()
     glCheckError();
 }
 
+
+QImage View3D::renderScreenShot(int width, int height)
+{
+    if (m_badOpenGL)
+        return QImage();
+
+    double dPR = getDevicePixelRatio();
```
c42f: As I understand things, this shouldn't be required, as `getDevicePixelRatio()` is related to the physical screen, not an offscreen buffer.

Instead, I'd suggest explicitly antialiasing, either by using `QOpenGLFramebufferObjectFormat`, or possibly more portably by just using a multiple of the actual width and height and downsampling on the CPU side (as you already have below with `scaleToWidth()`)

c42f: That is, just keep what we have, but get `dPR` from a setting or a fixed supersampling (eg, dPR = 2).  Probably want to rename it as such.

src/render/View3D.cpp
```diff
@@ -498,6 +514,42 @@ void View3D::paintGL()
...
+
+    // Note: the reason for the following two lines and the subsequent downsampling of the
+    //       resulting image is to make the radius of rendered points consistent between the
+    //       gui and screenshot on displas with dPR != 1.
+    width *= dPR;
+    height *= dPR;
+
+    QOpenGLFramebufferObject framebuffer(QSize(width, height), QOpenGLFramebufferObject::Depth);
+    framebuffer.bind();
+    resizeViewport(width, height);
```
c42f: This ends up calling `m_camera.setViewport()` which is wired into `restartRender()` via the `projectionChanged()` signal, so will probably force an unnecessary GUI redraw.  Will need some disentangling - perhaps just copying the required `glViewport()` stuff in here?  Really it's not a resize, so calling resize is a bit confusing anyway.

c42f:
Ok, I've had a look over things.  Generally looks good, other than some remaining entanglement between GUI state and offscreen rendering, plus the Qt5 issue.

src/render/View3D.cpp
landersson: The reason for scaling up the offscreen buffer by dPR and subsequently downscaling the resulting image by the same factor was just a workaround to make points rendered into the offscreen buffer the same size as points in the gui when rendered on my Retina Macbook (dPR=2). If rendering to the offscreen buffer at the same resolution as the gui window, the points in the screenshot end up being twice as large as those on the screen. I made a short attempt at figuring out exactly why this was happening but had to get on with other things. Any idea why this would happen?

c42f: Ah, of course!

The problem here is that in the vertex shader `gl_PointSize` must be specified in pixels, so the shader needs to know the size of the viewport.  This is communicated with the `pointPixelScale` uniform variable (see `View3d::drawPoints()`).  Just looking at that code, it currently assumes `dPR` should be multiplied in, and grabs information from the camera directly.  This is another thing which should be disentangled to make it possible to cleanly render to an offscreen buffer: everything about the camera and view should be passed along explicitly in `transState`.  Luckily transState already has `viewSize`, so you can probably just use something like

```C++
prog.setUniformValue("pointPixelScale", (GLfloat)(0.5*transState.viewSize.x*transState.projMatrix[0][0]));
```

landersson:
Just trying to use this to get some rendering done by repeatedly issuing the "-viewposition" and "-screenshot" commands from a python script, but it's pretty slow, I guess mainly because both of these commands will issue an unnecessary, complete gui redraw... I guess something that would solve the problem for me would be to send position and view angles as 6, probably optional, numeric parameters to the screenshot command.

c42f:
How many points do you have?  There's a few things which can cause it to be slow:
* Time to draw a complete frame.  If this is 1e8 points or so this could be several seconds.
* IPC overhead due to opening a new socket connection for each request.  This can introduce a quite a bit of latency (perhaps 50ms to open the socket?  Can't remember)
* As you said, forcing GUI redraws could make things twice as expensive.

The solution will be quite different depending on what the problem is.

landersson:
My biggest cloud is about 5e8 points. It's currently decimated by 3, but I was planning to use full density when rendering the final sequence. So yes, I guess it's not too surprising that it'll take a while to render all those points. The IPC overhead should be negligible in my case. I'll try to change things so an off screen render doesn't trigger a GUI redraw. Btw, just noticed that the -maxpoints command doesn't seem to work. It doesn't update the FileLoader object's internal max count setting which get's set when the FileLoader is instantiated.

c42f:
Thanks for pointing out the problem with `-maxpoints`, I've got a mostly-written patch for that which I'll try to get committed in the next couple of days.

landersson:
No problem. I'm still planning to get this pull request into a mergeable state at some point btw, but it might be a while until I have enough time to do it properly.

jkwashbourne:
Im guess this PR is abandoned? Is there any thought towards implementing the off screen anti-aliased screen shot command? Thanks much for your time.

PR #100, opened 2016-04-02
Suppress clang compiler warnings - register variables

nigels-com:


c42f:
Merged, thanks.

nigels-com:
warnings

PR #101, opened 2016-04-02
More compiler warnings

nigels-com:
gcc and clang compiler warnings for gcc 4.6, 4.7, 4.8, 4.9 and clang 3.6.0

warnings2

The remaining one seems harmless enough:

```
HCloudView.cpp:142, GNU Make + GNU C Compiler (gcc), Priority: Normal
function 'drawBounds' is not needed and will not be emitted [-Wunneeded-internal-declaration]
```


c42f:
Great, thanks :)

PR #102, opened 2016-05-06
Coverity fixes

nigels-com:
See also: https://scan.coverity.com/projects/nigels-com-displaz

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -255,8 +255,10 @@ int main(int argc, char *argv[])
             std::cout.write(msg.data(), msg.length());
             std::cout << "\n";
         }
-        catch (DisplazError & e) // receiveMessage can time out
+        catch (DisplazError & e)
         {
+            std::cerr << "ERROR: QUERY_CURSOR message timed out waiting for a response.\n";
+            return EXIT_FAILURE;
         }
     }
     if (maxPointCount > 0)
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -248,10 +248,16 @@ int main(int argc, char *argv[])
     }
     if (queryCursor)
     {
-        channel->sendMessage("QUERY_CURSOR");
-        QByteArray msg = channel->receiveMessage();
-        std::cout.write(msg.data(), msg.length());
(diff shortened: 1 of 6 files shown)


src/render/PointArray.cpp
```diff
@@ -154,8 +154,8 @@ static OctreeNode* makeTree(int depth, size_t* inds,
         return node;
     }
     // Partition points into the 8 child nodes
-    size_t* childRanges[9];
-    multi_partition(beginPtr, endPtr, OctreeChildIdx(P, center), childRanges+1, 8);
+    std::vector<size_t *> childRanges(9, NULL);
```
c42f: I assume this is about initialization, can we get away with `size_t* childRanges[9] = {0};` instead?  Seems a bit unfortunate to hit the allocator for this small fixed size array.

src/main.cpp
```diff
@@ -248,10 +248,16 @@ int main(int argc, char *argv[])
...
-        std::cout << "\n";
+        try
+        {
+            channel->sendMessage("QUERY_CURSOR");
+            QByteArray msg = channel->receiveMessage();
+            std::cout.write(msg.data(), msg.length());
+            std::cout << "\n";
+        }
+        catch (DisplazError & e) // receiveMessage can time out
+        {
```
c42f: Looks good.  Let's also log the error to `std::cerr` here, to give the user some sort of feedback.

src/render/PointArray.cpp
nigels-com: Yes, that would work.  `std::array` is an option too, assuming C++11

```
std::array<size_t *, 9> childRanges;
childRanges.fill(NULL);
````

which is arguably a bit clearer than the alternatives.

c42f: Should be safe to assume std::array is available.  Personally I'm perfectly happy with a good old C array, provided you don't need the type preserved as it's passed around.

nigels-com:
Updated, ought to be good to go.

c42f:
Cheers, thanks

PR #104, opened 2016-05-06
Update Catch unit test framework to version 1.5.0

nigels-com:
Updating catch.hpp to 1.5.0 seems to add as many Coverity defects as it resolves.
One manual defect resolution here to avoid copying parameters that would usually be const references.


c42f:
Merged, thanks.

Issue #87, opened 2016-01-07
Build failure on ubuntu-15.10


c42f:
The build fails on ubuntu 15.10, with qt headers complaining about `-fPIE`, which overrides `-fPIC` in the compile flags.  It's not very clear where the `-fPIE` is coming from (possibly cmake?)

juliohm:
@c42f I have the exact same issue on Arch Linux. Is there any workaround?

This actually may be an issue with the Qt 5 CMake modules, I also couldn't figure out where the -fPIE is coming from.

juliohm:
Google search returned a few interesting links:

* `Qt 5.4.2` important behavior changes: http://code.qt.io/cgit/qt/qtbase.git/plain/dist/changes-5.4.2
* Another attempt to fix it: https://github.com/Cockatrice/Cockatrice/commit/9cfb09cd5b25a01fb142d425ba70b77b89f65a09

I tried both, but they didn't work. I am using `GCC 5.3.0` and `CMake 3.4.2`.

c42f:
There's a workaround, but it's pretty disgusting ;-)

Build displaz according to the instructions, but replace the make step with 

```sh
make VERBOSE=1
```

This will spit out full information about each build step.  When you get to the failing step (IIRC it's a linker step), just copy the compile command which fails, remove the offending -fPIE and run it by hand.  The rest of the build should then go through, but if not, rinse and repeat...

Come to think of it, you could probably munge the makefiles using grep and sed which would make things slightly less manual.  How about trying something like the following from inside the build directory:

```sh
sed -i -e 's/\<fPIE\>/fPIC/g' $(grep -l -R -e '-fPIE')
```

then run make as usual.  Be aware that every time cmake is run you'll need to repeat this step before running make.

juliohm:
That is indeed a hack :blush:. For me the error happens when trying to compile `typespec.cpp`, it includes the header `glutil.h` which apparently include Qt headers.

After I remove `-fPIE` for this file, the build is fine. I opened `displaz-gui` and got this log:

![screenshot from 2016-01-30 07-39-18](https://cloud.githubusercontent.com/assets/3345261/12696653/01fda050-c725-11e5-9699-31608f5093e7.png)

Is everything ready for use or I need to install other missing dependencies?

juliohm:
It is fine, I had to go all the way through and install `make install` to have the `share` files in place. The screenshot above only shows the error because I tried to run from the build directory.

c42f:
[...]

PR #107, opened 2016-05-11
Force -fPIC for Ubuntu 16.04 Qt5 build

nigels-com:
This is a bit crude, but gets the job done on Ubuntu 16.04.
See also: https://github.com/c42f/displaz/issues/87


c42f:
Hah, the low tech fix. I merged it, it's certainly better than the build just not working :)

Issue #92, opened 2016-02-04
Hookable events for scripting interfaces


c42f:
For automating displaz from julia/python/etc it'd be really useful to be able to register keyboard/mouse event hooks from the scripting language and have data returned on each event.  This would make it possible to round trip from scripting language to form simple 3D UIs while prototyping algorithms.

Usage sketch:

1. Scripting language sends a hook command down the socket via the displaz CLI, eg `displaz -hookevent key:E cursorposition`, and sets up a task/thread blocking on reading the displaz process stdout.
2. displaz process waits on the socket
3. When displaz-gui reads the E key, it responds with the cursor position, which is read by displaz and sent to stdout
4. Script reads a line from stdout with the desired message, and responds by sending extra datasets back to displaz-gui

The hook in displaz-gui should disappear when the displaz CLI process exits.

The model above is pretty simple, but may be a bit command line centric and doesn't seem extensible to complex messages.  Possibly might want to consider a proper serialization format for socket messages rather than the ad hoc line-based message format.

c42f:
Additional useful hook event:  When user clicks on a point, return (dataset name, vertex index, vertex attributes) - more or less what currently goes in the log

landersson:
I nearly started writing something along these lines a little while ago when I wanted displaz-gui events to trigger actions in my python script. The way I planned to do it was to somehow give displaz-gui an ip+port to log events to, and have my python script open up a simple tcp server to record those. Your approach might be more straightforward. I ended up doing something less user friendly to get the task at hand done, but being able to pass displaz events back to a script is something I'd find quite useful.

c42f:
Internally displaz does something very much like that for its IPC, but with a local unix domain socket rather than a network socket.  For security reasons, I don't particularly want to harden displaz against external programs attacking it via the network.

Scripting language bindings can talk directly down the IPC socket in principle, but then each socket interface has to handle a few annoying things:

* If an existing displaz instance isn't found (nothing answers on the socket), either fail, or figure out how to open one.
* Deal with async socket IO; this involves varying levels of badness depending on the scripting language.
* The socket API isn't currently public.  We'd need a less hacky serialization format before making it so (json? msgpack?)

This is not to say I wouldn't consider a direct socket connection in the future, but it does come with some caveats which are currently handled in one place (ie, the displaz command line interface C++ code).

janzill:
I thought the pull request message would show up here but it doesn't, so here's a copy to discuss, sorry for the duplicate: 
[...]

PR #108, opened 2016-06-02
Hookable events from script (issue #92)

janzill:
I had a crack at issue #92 and implemented a simple CLI protocol along the lines of your outline, Chris. There are a couple of things I'm not entirely sure about:
 - Is the CLI supposed to wait for stdin after a single message is sent from the GUI, or would it be better to keep the hook active while no input is incoming (i.e. getting multiple events back)?
 - What kind of information is the script supposed to communicate back to the GUI, and what does the GUI do with it?
 - Do you want to be able to specify a key and event message to be returned to the CLI?

 Regarding the first point, the CLI is currently sending a single message to stdout and then waits for input from the script. If displaz would create another thread for the input and wait until it receives input, it would be possible to get as many events as desired with a single hook (the output would have to be blocked as soon as input starts).
 Regarding the second point, if there are complex messages to be sent back and forth a serialization of the socket communication messages would make the implementation easier as you mentioned.
 Regarding the third point, an event handler in a separate class might make sense here (this would also make View3D a bit smaller).

For now I implemented a single option (-hook) without arguments, and three events with associated behaviour (for reference: key:E returns coordinates of cursor position; right click+Control (command on Mac) returns position coordinates of closest point to click; right click+Shift returns more information on closest point to click ).  A combination of event handler and socket communication would make this easily extendable.


c42f:
Hi Jan, this looks like a great start.  I'll have a closer look and play with it as soon as I get a chance.

Regarding your questions -

> Is the CLI supposed to wait for stdin after a single message is sent from the GUI, or would it be better to keep the hook active while no input is incoming (i.e. getting multiple events back)?

I think the only robust option is for the scripting layer to get a stream of events from the displaz stdout.  To talk back to displaz, the simplest thing is for the script to just use the normal displaz command line with a separate displaz process.  Of course, that's not the most efficient thing, and an extra step would be to allow multiple commands to be sent down the displaz stdin (this could be a new instance of displaz though, it doesn't need to be the one on which hooked events are listened to).

With all this in mind, hooked events aren't a property of the displaz GUI itself, but a property of a particular IPC connection.

> What kind of information is the script supposed to communicate back to the GUI, and what does the GUI do with it?

The script should be able to do all the things it usually does, but since you can fire up a separate displaz command line instance to communicate with the same GUI this is pretty straightforward, if slightly less efficient in principle.

> Do you want to be able to specify a key and event message to be returned to the CLI?

I'll have to think a bit more about which events we want to communicate and how.  At a basic level it'd probably be useful to just be able to specify a set of keys (+modifiers) to listen for, and to echo that back to the command line each time an event happens.

janzill:
Ok, that makes it clearer, thanks. I had time on the weekend and worked on it, I tried to put all the hook-specific stuff into the IpcChannel class (at the cost of using sender() again, as well as the parent of IpcChannel explicitly). With the current implementation you can define one specific key set per displaz instance (one key plus one modifier) and return either the current cursor position or the key-shortcut itself.

janzill:
Apparently converting between stdstrings and QByteArras/QStrings was introduced in Qt5.4, which seems to be the reason why one of the tests failed. I'll check for backwards-compatible solutions tomorrow, it's getting late.

c42f:
I just spent a bit of time compiling and fiddling with this, it's very cool :-)  The `std::string` stuff was fairly easy to work around - I'd suggest avoiding std::string in Qt heavy code such as `IpcChannel`, and just using `QByteArray` and `QString`.  The only gotcha is that we need to keep in mind the character encoding of non-ascii characters on the socket should be consistent.  The convention I want to use is that all socket strings should be encoded as UTF-8.  (On linux, command line arguments should already be UTF-8, though on windows I don't think I've handled it correctly yet.)

Here's a few API things I think should be improved:
* The hook times out after 10s.  I think this should be set to -1 instead to avoid timing out at all.  The idea is that you'd subscribe to a stream indefinitely (see example below)
* I think the `-hookremove` command is unnecessary.  The way to remove a hook should probably be to simply close the `displaz -hook ...` process.  This could be done by abandoning the socket, or more nicely by catching an appropriate signal and closing the socket cleanly.  It really depends on which facilities are available in scripting languages as to what will be the best here.
* Multiple hook events should be possible though not necessarily in this pull request, eg `displaz -hook CTRL u key -hook CTRL y cursor ...`.  You'd probably need a callback to handle the multiple -hook arguments (see thirdparty/argparse.h, in particular the `%@` option.)
* For multiple events, the event type would need to come first on the command line so that events can be distinguished.
* The syntax for modifiers should be a bit more standard and in a single argument.  I'd prefer something like displayed in the GUI menus, eg `-hook Ctrl+U key`.  This may need a little parsing by splitting on the `+`, but that's pretty easy (or maybe Qt can do this for us?)

So for possible displaz command line usage I'm imagining the following bash script might eventually work:

```bash
displaz -hook Ctrl+U key -hook Ctrl+U cursor | \
while read event_type args ; do
    case $event_type in
        key)
            echo "Got key event $args"
            ;;
        cursor)
            echo "Got cursor position $args"
            ;;
    esac
done
```

Here's a bash script which actually does work with your current code (after fixing the timeout issue), and places a tetrahedron at the current position of the cursor whenever you press Ctrl+U.  This is the kind of fairly simple event loop design I'm hoping will be nice in other scripting languages (python, julia, matlab), but I haven't tried doing it there yet. 

testhook.sh
```bash
#!/bin/bash

set -u

calc() {
    echo "$@" | bc
}

make_tetrahedron() {
    x=$1
    y=$2
    z=$3
cat <<EOF
ply
format ascii 1.0
comment Simple tetrahedron
element vertex 4
property float x
property float y
property float z
property float r
property float g
property float b
element face 4
property list int int vertex_index
end_header
$(calc $x +  1.0) $(calc $y +  1.0) $(calc $z +  1.0)  1 0.8 0.8
$(calc $x +  1.0) $(calc $y + -1.0) $(calc $z + -1.0)  1 0.8 0.8
$(calc $x + -1.0) $(calc $y +  1.0) $(calc $z + -1.0)  1 0.8 0.8
$(calc $x + -1.0) $(calc $y + -1.0) $(calc $z +  1.0)  1 0.8 0.8
3 0 2 1
3 0 3 2
3 0 1 3
3 1 2 3
EOF
}

displaz -hook U CTRL cursor | \
while read x y z ; do
    echo $x $y $z
    temp_tet=$(mktemp tet_XXXXXXXXX.ply)
    make_tetrahedron $x $y $z > $temp_tet
    displaz -script -add -rmtemp -label Foo $temp_tet
done
```

janzill:
Thanks for the scripts and the list of improvements, I've tweaked a couple of things and multiple hook events from the same instance should now be working. I've also removed std::strings and the tests have passed. Regarding the other points:
> The hook times out after 10s. I think this should be set to -1 instead to avoid timing out at all. The idea is that you'd subscribe to a stream indefinitely (see example below)

I've set it to -1.

>I think the -hookremove command is unnecessary. The way to remove a hook should probably be to simply close the displaz -hook ... process. This could be done by abandoning the socket, or more nicely by catching an appropriate signal and closing the socket cleanly. It really depends on which facilities are available in scripting languages as to what will be the best here.

Ok, I've removed the -hookremove command, but haven't added any code to remove a hook cleanly. One possible solution might be to have std in listening on another thread.

>Multiple hook events should be possible though not necessarily in this pull request, eg displaz -hook CTRL u key -hook CTRL y cursor .... You'd probably need a callback to handle the multiple -hook arguments (see thirdparty/argparse.h, in particular the %@ option.) For multiple events, the event type would need to come first on the command line so that events can be distinguished.

Thanks for the tip, it's implemented like that now.

>The syntax for modifiers should be a bit more standard and in a single argument. I'd prefer something like displayed in the GUI menus, eg -hook Ctrl+U key. This may need a little parsing by splitting on the +, but that's pretty easy (or maybe Qt can do this for us?)

Qt does this, you can now specify a key set via Ctrl+U (not case sensitive). The modifier is optional but has to go first.


Two possible improvements, both related to implementing an event handler for key events in the GUI: 
- At the moment, the same key combination can't be active more than once at the same time, if it is there is no error thrown and the event won't be handled. The connected signal is activated(), for ambiguous events activatedAmbiguously() is emitted but currently not connected. 
- If one of the shortcuts that are already taken by the main window is specified (like ctrl+o or ctrl+q), QAction throws an error and nothing happens.

c42f:
Uh oh, something bad has happened to the diffs, looks like whitespace problems - you've now got windows file endings for most of the modified files.  It's pretty hard to review until that's fixed, can you change back to unix file endings (ideally by editing the offending commit to limit code churn). I've also got a couple of other pedantic whitespace requests: could you make sure all your tabs are converted to four spaces and kill off whitespace added at the ends of lines.  All this just makes things much more consistent between editors.

One thing I forgot to mention is that the test script doesn't work for me without adding a `std::cout.flush()` after reading each `recieveMessage()`.  Without this, the bash script doesn't get any events until the stdout buffer is full.

I'll have a closer look at the code after the whitespace problems are sorted out.

Regarding improvements, it makes sense to me to allow multiple scripts to subscribe to the same events, so having actions activated ambiguously seems reasonable.  Another alternative might be to have a central set of `QShortcut`s, with their signals connected to multiple slots.  I haven't looked closely at how you've hooked this up yet, but as an overall comment I'm not sure `IpcChannel` should be modified here - it's a generic tool for sending and receiving whole messages down a socket, and should probably stay that way.  I think it would be better to add a new class which manages a set of hooks and dispatches the resulting events down the socket via a given `IpcChannel`.  Hmm... `IpcEventDispatcher`?

I'm not entirely sure what to do about events which are already taken by the main window, but I'd be happy to just say these aren't available for now.

c42f:
Another thought.  I think it'll be useful/necessary to have the `cursor` event show the key which triggered it.  Actually these aren't so much *cursor* events as keyboard shortcut events which happen to have the 3D cursor location as a payload.  We should probably just reflect the Qt event class hierarchy with the event types, so that mouse events and other stuff could be added in the future without much disruption.  So perhaps hook usage should look a bit more like `-hook event_spec payload_spec`, eg

```
$ displaz -hook key:Ctrl+g cursor -hook key:Ctrl+f null
```

with resulting stream something like `event_spec payload_type payload_info`:
```
key:Ctrl+g cursor 1 2 3
key:Ctrl+f null
```

I hope this option might be a bit more future proof.

c42f:
By the way I'm quite excited by this, I think it'll enable some really cool and useful things to be built very quickly :-)

janzill:
Sorry for the formatting mess, I cleaned it up and amended the commit.

 >Another thought. I think it'll be useful/necessary to have the cursor event show the key which triggered it. Actually these aren't so much cursor events as keyboard shortcut events which happen to have the 3D cursor location as a payload. We should probably just reflect the Qt event class hierarchy with the event types, so that mouse events and other stuff could be added in the future without much disruption. So perhaps hook usage should look a bit more like -hook event_spec payload_spec

Ok, that makes sense. I'll have a look at this as well as implementing something like an IpcEventDispatcher class (including ambiguous shortcut handling) on the weekend, I'm pretty busy with finishing off a publication that has to be sent off by the end of the week.

c42f:
> Sorry for the formatting mess, I cleaned it up and amended the commit.

Thanks a lot for cleaning it up.  I just noticed that the file modes have also changed for some reason - also worth reverting.

> I'm pretty busy with finishing off a publication that has to be sent off by the end of the week.

No problems, don't let me distract you too much.  I think we're getting close with this now.

janzill:
>Thanks a lot for cleaning it up. I just noticed that the file modes have also changed for some reason - also worth reverting.

Ok, incorporated (again forcefully amended, I'm assuming I'm the only one working on it at the moment). I'm not sure what happened there to be honest, I'll try to find out so it doesn't happen again. 

> No problems, don't let me distract you too much. I think we're getting close with this now.

I actually enjoy working on this, it's more fun than polishing a manuscript.

c42f:
> more fun than polishing a manuscript

`<laugh class="evil">Heh heh heh</laugh>`

c42f:
(yup, that was pretty much the story of my PhD)

janzill:
> <laugh class="evil">Heh heh heh</laugh>

5 1/2 weeks and I will be able to laugh about this too :-)

janzill:
Ok, I've worked on this some more.

>Regarding improvements, it makes sense to me to allow multiple scripts to subscribe to the same events, so having actions activated ambiguously seems reasonable. Another alternative might be to have a central set of QShortcuts, with their signals connected to multiple slots. I haven't looked closely at how you've hooked this up yet, but as an overall comment I'm not sure IpcChannel should be modified here - it's a generic tool for sending and receiving whole messages down a socket, and should probably stay that way. I think it would be better to add a new class which manages a set of hooks and dispatches the resulting events down the socket via a given IpcChannel. Hmm... IpcEventDispatcher?

I've implemented a two new classes with an eye on making this feature future proof. 
The first one (hookEvent) is meant to store a list of custom events and register them. For now this is all implemented via QShortcuts and taken care of by Qt, but QShortcut does not support mouse events. I'll have a look at custom events later tonight to see if I can extend this.
The second one (IpcEventDispatcher) is meant to dispatch events down a the socket. Every hook specifier for every channel has its own instance, this way channels are discernible. It might be possible to make this more efficient by implementing a signal mapper and only having one instance. One problem this implementation causes is that if one specific displaz instance signs up to the same hook event more than once (and therefore messages are sent to the same channel simultaneously), one or more messages might get lost (one always comes through, the others are pretty random). This can be avoided by adding a millisleep(1) at the end of IpcChannel's sendMessage, but I'm not sure if this is appropriate.

> Another thought. I think it'll be useful/necessary to have the cursor event show the key which triggered it. Actually these aren't so much cursor events as keyboard shortcut events which happen to have the 3D cursor location as a payload. We should probably just reflect the Qt event class hierarchy with the event types, so that mouse events and other stuff could be added in the future without much disruption. So perhaps hook usage should look a bit more like -hook event_spec payload_spec, eg

At the moment, if the payload specifier is not 'cursor', the program just echos the given option.

Ok, dinner time and then I'll have a look at adding mouse events.

janzill:
I don't think adding mouse events is as straightforward as I thought. 
I've incorporated your suggestion on adding an event keyword before the specifier ('key:ctrl+u') by simply removing everything before a colon (i.e. the old syntax of -hook ctrl+u ...still works), and the output for everything but payload specifier cursor is 'null'. Also, key sequences are now disabled when no channel is listening.

c42f:
Great, I'll have a more detailed look through the code now.

src/hookEvent.h
```diff
@@ -0,0 +1,32 @@
...
+#ifndef DISPLAZ_HOOKEVENT_H_INCLUDED
+#define DISPLAZ_HOOKEVENT_H_INCLUDED
+
+#include <QByteArray>
+#include <QKeySequence>
+#include <QShortcut>
+
+/// Class for registering custom hook events. Implemented via QShortCut,
+/// i.e. for QKeySequences only.
+class hookEvent : public QObject
```
c42f: Naming: this should be `HookEvent` in keeping with naming of the other classes.  For new classes, I'm also trying to follow the convention that the file is named exactly after the class, so the files should be `HookEvent.{h,cpp}`

c42f:
I'm not too worried about supporting mouse events for now.  Let's get this polished up and committed, it's already going to be a great feature.

src/hookEvent.h
c42f: Actually, before you go renaming, it looks like `HookEvent` isn't right either in the current design: It's actually the full list of hook events, so it should be called `HookEvents` if nothing else.  Let me mull over the design a little.

src/main.cpp
```diff
@@ -276,6 +292,32 @@ int main(int argc, char *argv[])
         channel->sendMessage(QByteArray("OPEN_SHADER\n") +
                              shaderName.c_str());
     }
+    if(hookSpecDef != "")
```
c42f: I think this would be (very slightly) more logical as `if (!hookPayload.empty())`

src/main.cpp
```diff
@@ -276,6 +292,32 @@ int main(int argc, char *argv[])
...
+    {
+        QByteArray msg;
+        QByteArray message = QByteArray("HOOK");
+        for(int i=0; i<hookPayload.size(); i++)
+            message = message + QByteArray("\n")
+                              + QByteArray(hookSpec[i].data(), (int)hookSpec[i].size()) + QByteArray("\n")
+                              + QByteArray(hookPayload[i].data(), (int)hookPayload[i].size());
```
c42f: Style nitpick: for consistency, please insert some braces here.

If the body of the loop (of if statement) is a single line, I prefer to leave the braces off, but when it starts to cover multiple lines I think it's best to add some.  (I'm probably not entirely consistent with this myself...)

src/hookEvent.h
```diff
@@ -0,0 +1,32 @@
...
+{
+    Q_OBJECT
+    public:
+        hookEvent(QObject* parent = NULL);
+
+        void setHookEvent(QByteArray hookSpec);
+        int getHookId() const { return latestHookIndex; }
```
c42f: I think these two methods should go together, with `setHookEvent` just returning the id.  Then you don't need to keep the `latestHookIndex` state at all.

c42f: Again, let me have a little more of a think about how things hang together here.  In particular I'm not yet sure about the integer ids.

src/IpcEventDispatcher.cpp
```diff
@@ -0,0 +1,30 @@
...
+// Use of this code is governed by the BSD-style license found in LICENSE.txt
+
+#include "IpcEventDispatcher.h"
+#include "IpcChannel.h"
+#include "PointViewerMainWindow.h"
+
+
+IpcEventDispatcher::IpcEventDispatcher(QObject* parent, PointViewerMainWindow* getPayload, 
+                                       QByteArray hookSpec, QByteArray hookPayload, int whichShortCut)
+    : QObject(parent),
```
c42f: This is an interesting design, making `IpcEventDispatcher` a child of the channel.  It's not what I had in mind, but from what I can see so far it seems like a pretty good idea.

In this design, the event dispatcher is more of an event formatting class, rather than explicitly dispatching the event to the underlying channel.  Perhaps `IpcEventFormatter` would make more sense?

src/IpcEventDispatcher.h
```diff
@@ -0,0 +1,40 @@
...
+class PointViewerMainWindow;
+
+/// Class for dispatching hook events and payloads to given IpcChannel
+///
+class IpcEventDispatcher : public QObject
+{
+    Q_OBJECT
+    public:
+        IpcEventDispatcher(QObject* parent = NULL, PointViewerMainWindow* getPayload = NULL,
+                           QByteArray = 0, QByteArray = 0, int = 0);
```
c42f: It's probably not a good idea to default all these arguments, because if some of them are ever null (particularly `getPayload`) you're likely to get a nice segfault :-)

src/hookEvent.h
c42f: Or how about `HookManager`, that might be even better?

janzill: <HookManager> describes it concisely I think.

src/hookEvent.h
janzill: Good idea, I'll change it if you are happy with keeping the integer ids.

src/IpcEventDispatcher.cpp
janzill: I made it a child of the main window at first, but thought this way Qt would take care of cleaning up during a gui session. The class is formatting the message and the instances keep the hook signals discernible, so IpcEventFormatter does make sense.

src/IpcEventDispatcher.h
janzill: I'll change it instantly :)

janzill:
Thanks for the feedback. Are you happy with the design in general? If so, I'm going to implement the suggested changes.

src/hookEvent.h
```diff
@@ -0,0 +1,32 @@
...
+        QVector<QShortcut*> m_shortCut;
```
c42f: Probably shouldn't be public.

c42f:
Sorry, have been a bit distracted, I only just got back to looking over this.

I had a look through, as noted I think some things could be named a bit more clearly, and there's a few places we might possibly be able to streamline the design by making better use of the signals and slots mechanism.  Let me try to summarize the pieces as I see them, and we'll see how we go.

Here's a bit of a dataflow graph of what's going on:

```
QKeyEvent
  |
  |
  |
  |--> QShortcut (Ctrl+A)
  |    |
  |    |
  |    |--> IpcEventFormatter.1 cursor_payload --> IpcChannel.1
  |    |
  |    |
  |    \--> IpcEventFormatter.2 null_payload -----\
  |                                               |
  |                                               |
  \--> QShortcut (Ctrl+B)                         |
       |                                          |
       |                                          |
       \--> IpcEventFormatter.1 cursor_payload ------>  IpcChannel.2
```

The event formatters should be deleted when the channels are deleted, likewise the shortcuts should be deleted (or disabled) when there's no formatters left for a given shortcut.  It looks like you've got both those sorted out, but you might be able to be a bit more direct about it rather than going through the integer ids.

I'll have more a think about this, but I wanted to get the above written down in a comment while I do so :-)

c42f:
The tricky thing is that a single `QShortcut` (or would a QAction with an associated shortcut be better?) is associated with multiple formatters, so managing the lifetime of the QShortcut is tricky since Qt connections won't keep the shortcuts objects alive - we've got to hold a reference to them elsewhere.  What you've got here seems pretty reasonable, right now I can't think of an alternative to the manual reference counting which is obviously simpler and cleaner.  (We could go with some `QWeakPointer`s inside `HookManager` and `QSharedPointer`s held by the formatters to make it automatic, but this just seems to increase the coupling and make things more complicated.)

src/hookEvent.h
janzill: Right, forgot to make it private and add an access member function after I was done experimenting.

c42f:
> this just seems to increase the coupling and make things more complicated

On the other hand, maybe not so much: the hook formatter already needs to keep an event id around - whether that id is an integer or a `QSharedPtr<QObject>`, is probably a bit immaterial.  I do wonder whether there's any way to avoid this coupling.  The problem being that a Qt connection won't keep an object alive by itself.

src/IpcEventDispatcher.cpp
c42f: > this way Qt would take care of cleaning up

Yes, this does make sense, I like it.

c42f: Or more appropriately, perhaps just `HookFormatter`?  It's not just the event which is formatted, it's the payload, etc.

Expanding on this, it would make sense to me if we had multiple hook formatter implementations, one for each type of hook payload.  For now I'd say that's an unnecessary extension, but it'd be easy to do later.

janzill:
Hm, I thought about this a bit on the weekend and couldn't find a simple alternative. I'll have a look at shared pointers.

src/IpcEventDispatcher.cpp
janzill: The hooks are registered and managed in HookManager, and the message corresponding to a hook event is formatted in HookFormatter, the dispatching being taken care of by simply connecting these objects with the appropriate channel via signals and slots. I'll change the names.

src/hookEvent.h
```diff
@@ -0,0 +1,32 @@
...
+
```
c42f: Having thought about it, I think you can put these three public things together into one.  Something roughly like

```C++
/// Connect events defined by `eventSpec` to the HookFormatter hookActivated() function.
void connectHook(QByteArray eventSpec, HookFormatter* formatter)
```

src/gui/PointViewerMainWindow.cpp
```diff
@@ -468,13 +474,49 @@ void PointViewerMainWindow::handleMessage(QByteArray message)
...
+        for(int i=1; i<commandTokens.count(); i+=2)
+        {
+            m_hookEvent->setHookEvent(commandTokens[i]);
+            int latestHookIndex = m_hookEvent->getHookId();
+            // new object for each channel and each payload, keeps 
+            // channels and payloads discernible without SignalMapper
+            IpcEventDispatcher* hookDispatch = new IpcEventDispatcher(channel, this, commandTokens[i], commandTokens[i+1], latestHookIndex);
+
+            connect(m_hookEvent->m_shortCut.at(latestHookIndex), SIGNAL(activated()),
+                    hookDispatch, SLOT(hookActivated()));
```
c42f: I think most of this should go into `HookManager`.

c42f:
To summarize:

* IpcEventDispatcher -> HookFormatter
* hookEvent -> HookManager
* HookManager public interface could be simplified to change the way signals are connected and event ids are passed around.  The list of shortcuts should be private.

Hopefully that covers most things, I've got to rush home now!

janzill:
Thanks for the review, I cleaned it up.

janzill:
I thought about the design a bit more (without having any better ideas I'm afraid) and noticed that there was a bug, the counter of signed up hooks in HookManager did not get decremented on disconnection of a hook.

src/HookManager.cpp
```diff
@@ -0,0 +1,64 @@
...
+        connect(m_event.at(numberEvents), SIGNAL(activated()), formatter, SLOT(hookActivated()));
+        formatter->setEventId(numberEvents);
+    }
+
+    connect(formatter, SIGNAL(hookRemoved(int)), this, SLOT(deactivateEvent(int)));
+}
+
+
+void HookManager::deactivateEvent(int eventId)
+{
```
c42f: It's probably worth checking that `eventId` is in the valid range of `m_hooksSignedUp` here.  If not, it's a programming error - log an error (to stderr?) and return I guess?

src/HookManager.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+
+void HookManager::connectHook(QByteArray eventSpec, HookFormatter* formatter)
+{
+    /// Remove keyword before key sequence, indicated by colon. At the moment,
+    /// everything up to (and including) the colon is removed and discarded, 
+    /// so not specifying 'key:' works as well.
+    QString specifier = QString::fromUtf8(eventSpec);
+    //specifier.count(QRegExp("[:]"));
+    specifier.replace(QRegExp(".*[:]"), "");
```
c42f: I think we should be prescriptive here and only accept a `"key:"` prefix.  Otherwise people will come to rely on the permissive API and their scripts will break later on.

src/HookManager.h
```diff
@@ -0,0 +1,33 @@
...
+
+        /// Connect events defined by `eventSpec` to the HookFormatter hookActivated() function.
+        void connectHook(QByteArray eventSpec, HookFormatter* formatter);
+
+    public slots:
+        void deactivateEvent(int whichEvent);
+
+    private:
+        QVector<QShortcut*> m_event;
+        QList<QKeySequence> m_eventSpec;
```
c42f: Probably better as `m_events` and `m_eventSpecs`

src/HookManager.cpp
```diff
@@ -0,0 +1,64 @@
...
+    QKeySequence requested = QKeySequence::fromString(specifier, QKeySequence::PortableText);
+
+    /// Go through all shortcuts, see if one exists, if not create it and connect it to HookFormatter
+    int numberEvents = m_event.count();
+    bool taken = false;
+    for(int i=0; i<numberEvents; i++)
+    {
+        if(m_eventSpec.at(i)==requested)
```
c42f: I'd suggest the unchecked `m_eventSpecs[i]` rather than using `at()`, for consistency with all the other code.  This goes for most uses of `at()` really.

The style I tend to follow is to check lengths explicitly in those places which need it, and avoid the overhead elsewhere.

c42f: That is, the error handling philosophy here is not so much to use exceptions to catch *programming* errors - I tend to use `asserts()` for that, if the logic seems particularly tricky.

src/gui/PointViewerMainWindow.cpp
```diff
@@ -468,13 +474,42 @@ void PointViewerMainWindow::handleMessage(QByteArray message)
...
+        IpcChannel* channel = dynamic_cast<IpcChannel*>(sender());
+        if (!channel)
+        {
+            qWarning() << "Signalling object not a IpcChannel!\n";
+            return;
+        }
+        for(int i=1; i<commandTokens.count(); i+=2)
+        {
+            HookFormatter* formatter = new HookFormatter(this, commandTokens[i], commandTokens[i+1], channel);
+            m_hookManager->connectHook(commandTokens[i], formatter);
```
c42f: Neat

c42f:
I think we're done here, very nice :-)

If I can bother you to clean up a few minor detail-related issues that would be great, then I think I'll merge this.

src/HookManager.cpp
janzill: I implemented this such that if the specifier is not of the form 'key:eventSpec', a message gets sent down the listening channel, and then the formatter is deleted, but the channel is still in use.

janzill:
Cool, it's done provided you are happy with the error handling for the event specifier.

c42f:
Yeah looks good.  I fixed a compiler warning and merged it.

PR #109, opened 2016-06-10
Replace labeled datasets

c42f:
Change the way that `-add` and `-clear` work to be a lot more helpful.

@JoshChristie this is the long-awaited change I keep mentioning to make over plotting more helpful by default.


JoshChristie:
Very nice, this will be useful! :+1:

Issue #99, opened 2016-04-01
Windows package broken


JoshChristie:
After installing the continuous windows build (https://displaz-data.s3.amazonaws.com/builds/ci/latest/displaz-win64.exe), I get the following error when I try to run the executable. 
![displaz](https://cloud.githubusercontent.com/assets/7004868/14196626/d39697f6-f809-11e5-9eb4-773547dd82bb.png)

nigels-com:
Yes, confirming the same problem here on Windows 7, 64-bit.  The workaround that seems to work for me is to use some Qt dlls from a previous install of displaz.

c42f:
Looks like this is due to Qt making more things into plugins so they can't be found automatically via `fixup_package()`.  Relevant links

http://stackoverflow.com/questions/31920698/how-to-deploy-qwindowsintegrationplugin-with-cmake

Looks like there's some relevant cmake song and dance in GCompris-qt 
https://github.com/bdoin/GCompris-qt/blob/e88ea239056c54410451cc5da660ad264dd7ae37/src/core/CMakeLists.txt#L96
https://github.com/bdoin/GCompris-qt/blob/a33ff0300ce352e7b727e0c8e39b6cb5ebd16947/cmake/qt_helper.cmake#L48

Ick :-(

c42f:
Another useful link

https://gitlab.kitware.com/kotu/MoleQueue-Qt5.0.2/blob/a1dcf93ed46a7c7a101ae448f912a3351679d3cf/cmake/DeployQt5.cmake

c42f:
A junky workaround for this looks something like

```
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 80378f5..06340d8 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -231,6 +231,9 @@ if (DISPLAZ_BUILD_DVOX)
     install(TARGETS dvox DESTINATION "${DISPLAZ_BIN_DIR}")
 endif()
 
+get_property(loc TARGET Qt5::QWindowsIntegrationPlugin PROPERTY LOCATION_RELEASE)
+install(FILES ${loc} DESTINATION bin/plugins)
+
 if(WIN32)
     # Install required qt libraries
     if (DISPLAZ_USE_QT4)
```

c42f:
Fixed, thanks to @evetion in #168

PR #110, opened 2016-06-14
Fix windows package

c42f:
An attempt to deal with #99, let's see how it goes on appveyor.

### DIFF
--- a/cmake/DeployQt5.cmake
+++ b/cmake/DeployQt5.cmake
@@ -1,0 +1,345 @@
+#.rst:
+# DeployQt5
+# ---------
+#
+# Functions to help assemble a standalone Qt5 executable.
+#
+# A collection of CMake utility functions useful for deploying Qt5
+# executables.
+#
+# The following functions are provided by this module:
+#
+# ::
+#
+#    write_qt5_conf
+#    resolve_qt5_paths
+#    fixup_qt5_executable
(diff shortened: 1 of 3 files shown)


cmake/DeployQt5.cmake
```diff
@@ -0,0 +1,373 @@
...
+      get_property(loc TARGET Qt5::QCocoaIntegrationPlugin
+        PROPERTY LOCATION_RELEASE)
+      install_qt5_plugin("${loc}" "${executable}" 0 installed_plugin_paths
+        "${plugins_dir}/platforms" "${component}")
+      list(APPEND libs ${installed_plugin_paths})
+    elseif(WIN32)
+      get_property(loc TARGET Qt5::QWindowsIntegrationPlugin
+        PROPERTY LOCATION_RELEASE)
+      install_qt5_plugin("${loc}" "${executable}" 0 installed_plugin_paths
+        "platforms" "${component}")
```
evetion: "platforms" here leads to an extra nested folder, can be removed

c42f: Thanks for the suggestion.  Does packaging work for you if this is changed?

(As you can tell, I lost steam on fixing this packaging issue, which is part of the reason displaz hasn't had a proper release in far too long.)

c42f:
Superseded by #168

Issue #86, opened 2015-12-13
Segfault after modifying dataset list


c42f:
To reproduce:

* Open two point files (test/single_point.ply will do)
* Open Data Sets tab
* Select second data set with mouse; press delete to remove it from the list
* Select first data set; displaz will now segfault :-(

According to gdb, the crash is inside the OpenGL driver, as called from `pointarray.cpp:721`.

c42f:
Seems like this is not entirely deterministic - selecting something other than the first in the list sometimes also causes a crash, and the procedure above doesn't always reproduce the bug.

landersson:
Hi Chris, 

I also just ran into exactly this problem while modifying displaz to add a new remote command for removing datasets. I'm running Ubuntu 14.04 with NVIDIA driver 352.79 using a GTX 670. Let me know if you figure out what's going on. My knowledge of modern OpenGL is probably not up to scratch for debugging this further.

Anyway, thanks for a very useful program and nicely hackable code... I've found displaz invaluable for visualizing and interacting with large point cloud data sets in combination with .ply vector data.

c42f:
I first noticed this right after the Qt5 port which makes me think it's an issue with datasets being accessed after deletion by the UI list code, or some such thing (potentially unrelated to OpenGL, but segfaulting there)...  Indeed it's a nasty and tricky bug to track down, and trying to reproduce it on my work computer just now failed :-(  Will try again later at home.

I'm glad you're finding displaz useful!  I plan to continue adding remote commands to make the UI as scriptable as possible, so removing data sets remotely is definitely within scope.  My current plan is first to make it unnecessary in many cases, by making the `-dataname` option replace an existing data set with the same name by default.  We'd also want a way to refer to datasets by a unique ID to allow for more complex manipulation (this may be the name, or it may just default to the name, I'm not sure.  See the dataset-id-interface branch where I started down the path of having a uuid per dataset.  This turned out to be a bit messy and rather too chatty over the socket, so I don't plan to merge it in the current state, but it's a potentially interesting point in the design space.)

c42f:
By the way, if you're planning to add new commands and you'd like to contribute them, please open an issue and we can discuss the best design.  In the future I'd like to officially document the socket protocol, but only after it's made into something which sucks a lot less ;-)  Something like a json or msgpack based command language would probably make more sense than the mess I currently have :)

landersson:
[...]

PR #113, opened 2016-07-01
Opengl buffer bind cleanup

c42f:
Finally get to the bottom of #86 - as expected a stale piece of OpenGL state, in this case it was the vertex array object which wasn't unbound.  Inside this VAO was referenced a vertex buffer object with a pointer to CPU memory which could become stale when a geometry object was deleted.  Eventually the driver would access this stale memory, causing a segfault in the driver.  Ugh!

@r-chris - note that VAOs should be carefully unbound or this kind of thing happens due to opengl global state leaking.  It's permissible to delete the VBOs used to pass data as long as the VAO is unbound first (forcing the VAO to keep the reference around internally).

### DIFF
--- a/src/render/Geometry.cpp
+++ b/src/render/Geometry.cpp
@@ -125,7 +125,6 @@ void Geometry::initializeBboxGL(unsigned int bboxShader)
             4,5, 5,6, 6,7, 7,4
     };
 
-
     // create VBA VBO for rendering ...
     GLuint bboxVertexArray;
     glGenVertexArrays(1, &bboxVertexArray);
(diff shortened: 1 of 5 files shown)


r-chris:
I vaguely remember having trouble after unbinding under Mac OS X, but there probably was some bug in my code at the time - I'll give this a try when I get around to rebuilding Displaz.

c42f:
Thanks.  I found that it's quite easy to get segfaults if you let OpenGL state leak from a function - the trick seems to be unbinding things aggressively once you're done with them.

PR #115, opened 2016-07-05
Fixes for non-RGBA textures

### LINKED ISSUE #114: Various texture problems
A couple of problems with the new texture stuff which somehow slipped through:

* Segfault if you use a greyscale png as a TextureFile
* Random rainbows if you use a normal RGB image without an alpha channel
* A random colour is used when a texture file was specified but not found at the given path.  A warning should be generated instead.

c42f:
* Use QGLWidget::convertToGLFormat() to produce a QImage with OpenGL
  GL_RGBA compatible bit layout rather than doing it manually.
* Warn when a texture is missing
* Remove info logging when loading textures

Fixes #114


nigels-com:
Yay! :-)

PR #119, opened 2016-07-21
Add center camera on given label feature

JoshChristie:
Adds a new parameter `--centeronlabel` which centers the camera on the given label name.

### DIFF
--- a/src/geometrycollection.h
+++ b/src/geometrycollection.h
@@ -33,6 +33,9 @@ class GeometryCollection : public QAbstractListModel
         /// Remove and unload all geometries whose filenames matched by given QRegExp object
         void unloadFiles(const QRegExp & filenameRegex);
 
+        /// Find the first index to a geometry with label matching the given pattern
+        QModelIndex findLabel(const QRegExp & labelPattern);
+
         // Following implemented from QAbstractListModel:
         virtual int rowCount(const QModelIndex & parent = QModelIndex()) const;
         virtual QVariant data(const QModelIndex & index, int role = Qt::DisplayRole) const;
--- a/src/gui/PointViewerMainWindow.cpp
+++ b/src/gui/PointViewerMainWindow.cpp
@@ -397,6 +397,20 @@ void PointViewerMainWindow::handleMessage(QByteArray message)
         }
         m_geometries->unloadFiles(regex);
     }
+    else if (commandTokens[0] == "SET_VIEW_LABEL")
+    {
+        QString regex_str = commandTokens[1];
+        QRegExp regex(regex_str, Qt::CaseSensitive, QRegExp::FixedString);
+        if (!regex.isValid())
+        {
+            g_logger.error("Invalid pattern in -unload command: '%s': %s",
+                           regex_str, regex.errorString());
+            return;
+        }
+        QModelIndex index = m_geometries->findLabel(regex);
+        if (index.isValid())
+            m_pointView->centerOnGeometry(index);
+    }
     else if (commandTokens[0] == "SET_VIEW_POSITION")
     {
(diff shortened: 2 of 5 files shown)


c42f:
Thanks Josh!

src/geometrycollection.h
```diff
@@ -33,6 +33,9 @@ class GeometryCollection : public QAbstractListModel
         /// Remove and unload all geometries whose filenames matched by given QRegExp object
         void unloadFiles(const QRegExp & filenameRegex);
 
+        /// Find QModelIndex matching the label
```
c42f: Let's be slightly more specific here: `Find the first index to a geometry with label matching the given pattern`

src/gui/PointViewerMainWindow.cpp
```diff
@@ -397,6 +397,18 @@ void PointViewerMainWindow::handleMessage(QByteArray message)
...
+    {
+        QString regex_str = commandTokens[1];
+        QRegExp regex(regex_str, Qt::CaseSensitive, QRegExp::FixedString);
+        if (!regex.isValid())
+        {
+            g_logger.error("Invalid pattern in -unload command: '%s': %s",
+                           regex_str, regex.errorString());
+            return;
+        }
+        m_pointView->centerOnGeometry(m_geometries->findLabel(regex));
```
c42f: What happens if `findLabel()` returns an invalid QModelIndex (may be tested with `isValid()`)?

src/main.cpp
```diff
@@ -99,6 +100,7 @@ int main(int argc, char *argv[])
         "-label %s",     &g_dataSetLabel, "Override the displayed dataset label for the next data file on the command line with the given name",
 
         "<SEPARATOR>", "\nInitial settings / remote commands:",
+        "-centeronlabel %s", &centerLabelName, "Center camera on the dataset with the given label name",
```
c42f: The ordering here is what the user sees on the command line `-help`.  Let's put this down after to `-viewradius` since it's essentially setting the view position and radius.

Could be worth renaming it to start with `-view`, though nothing terribly convincing comes to mind... do you like any of `-viewatgeometry` `-viewtogeometry` `-viewtoward` `-viewlabel`?

src/render/View3D.cpp
```diff
@@ -212,7 +212,7 @@ void View3D::centerOnGeometry(const QModelIndex& index)
     m_cursorPos = geom.centroid();
     m_camera.setCenter(m_cursorPos);
     double diag = (geom.boundingBox().max - geom.boundingBox().min).length();
-    m_camera.setEyeToCenterDistance(diag*0.7);
+    m_camera.setEyeToCenterDistance(diag*0.7 + 0.01);
```
c42f: Thanks for fixing this!  There's no obvious choice for the epsilon, I guess we'll go with this and see what happens :-)

src/main.cpp
JoshChristie: I kind of like `-viewlabel` due to its simplicity.

src/gui/PointViewerMainWindow.cpp
JoshChristie: Was this what you had in mind?
```julia
QModelIndex index = m_geometries->findLabel(regex);
if (index.isValid())
    m_pointView->centerOnGeometry(index);
```

JoshChristie:
Done with the recommended changes :) thanks!

src/main.cpp
c42f: Cool, looks good!

src/gui/PointViewerMainWindow.cpp
c42f: Exactly :)

PR #121, opened 2016-08-26
RFC: Changed mouse controls

andyferris:
I'd like some feedback on some possible UI modifications I'm playing with in displaz. Some things I'm considering when designing this include the fact that its difficult to press the middle mouse button on modern mice, a UI which is consistent with possible future touch controls, and an analogue/continuous method for translating the centre of view.

Specific modifications are:

* Double click for select point and recenter
* Right click becomes pan. In standard mode it pans in the x-y plane. In
  trackball mode it pans in the camera/image plane. Might be nice for
  touchscreen interface with two finger pan, one finger rotate, and pinch-to-zoom?
* Seperate codepath for mouse wheel (same UI effect though)

For panning, the choices I've made specifically ignore the point cloud *and* the location on the screen the user initiates the drag from. We still need controls (maybe keyboard bindings) for z displacements.

Also, possibly standard WASD controls would obviate the need for pan-by-mouse mode and be more familiar to gamers. I think we'd want Q/Z or something for going up/down.

### DIFF
--- a/src/gui/InteractiveCamera.h
+++ b/src/gui/InteractiveCamera.h
@@ -218,20 +218,75 @@ class InteractiveCamera : public QObject
             m_trackballInteraction = trackballInteraction;
         }
 
+        /// Zoom the camera using a drag of the mouse
+        ///
+        /// The previous and current positions of the mouse during the move are
+        /// given by prevPos and currPos. The camera position is zoomed in
+        /// toward the center.
+        void mouseZoom(QPoint prevPos, QPoint currPos)
+        {
+            // exponential zooming gives scale-independent sensitivity
+            qreal dy = qreal(currPos.y() - prevPos.y())/m_viewport.height();
+            const qreal zoomSpeed = 3.0f;
+            m_dist *= std::exp(zoomSpeed*dy);
+            emit viewChanged();
+        }
+
+
         /// Move the camera using a drag of the mouse.
         ///
         /// The previous and current positions of the mouse during the move are
         /// given by prevPos and currPos.  By default this rotates the camera
-        /// around the center, but if zoom is true, the camera position is
-        /// zoomed in toward the center instead.
-        void mouseDrag(QPoint prevPos, QPoint currPos, bool zoom = false)
+        /// around the center, but if pan is true, the camera position is
+        /// panned in the x-y plane (or image plane in trackball mode) instead.
+        void mouseDrag(QPoint prevPos, QPoint currPos, bool pan = false)
         {
-            if(zoom)
+            if(pan)
             {
                 // exponential zooming gives scale-independent sensitivity
+                //qreal dy = qreal(currPos.y() - prevPos.y())/m_viewport.height();
+                //const qreal zoomSpeed = 3.0f;
+                //m_dist *= std::exp(zoomSpeed*dy);
+
+                qreal dx = qreal(currPos.x() - prevPos.x())/m_viewport.height();
                 qreal dy = qreal(currPos.y() - prevPos.y())/m_viewport.height();
-                const qreal zoomSpeed = 3.0f;
-                m_dist *= std::exp(zoomSpeed*dy);
+
+                if (m_trackballInteraction)
+                {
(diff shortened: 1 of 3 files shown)


c42f:
Hi Andy, thanks for looking at this.  The middle mouse button problem is definitely something that needs improving.

After playing with this for ten minutes, I'm afraid it drives me completely nuts though!  Impressions:

* For my typical use case, I'm inspecting an object in the point cloud.  I want to focus on that object in particular, which is why a single middle click works well for me: it's relatively precise and efficient, it usually doesn't leave you guessing about how the camera is going to behave.
* Zooming is just as important as panning; if you want to support two button mice or touchpads, we also can't rely on the mouse wheel to zoom.
* Having something graphical to indicate the 3D camera centre is quite useful, as it provides expectation of how rotation will work (and also panning in the patch here).  I tried decoupling these a while back and reverted it after a month or so as I didn't like the result.  If we want to have a 3D cursor for other things perhaps it would be worth trying multiple cursors (in different styles?).
* Double click to centre seems pretty workable.
* Having two quite ways to move the centre of the 3D camera seems like a bad thing.
* Having the geometry stick to the mouse as it moves might improve things.

So for now, I'm still not convinced that panning this way is going to be better.

On the multi touch front, perhaps that system should just be a bit different from the mouse-based one.  vrui has some interesting multi-touch like interaction: https://www.youtube.com/watch?v=cyoJKbzqpZA

andyferris:
> After playing with this for ten minutes, I'm afraid it drives me completely nuts though!

Yes I agree that it definitely needs improving significantly!

> For my typical use case, I'm inspecting an object in the point cloud. I want to focus on that object in particular, which is why a single middle click works well for me: it's relatively precise and efficient, it usually doesn't leave you guessing about how the camera is going to behave.

I'd add that the double click was more precise in my experimentation - the cursor was less likely to move when I double left-click as opposed to middle click.

> Zooming is just as important as panning; if you want to support two button mice or touchpads, we also can't rely on the mouse wheel to zoom.

Probably a good point, but I'd wager there are more mice with a scroll wheel than a good middle button. Touchpads, apple magic mice, etc all support two finger scroll. Does that work currently?

> Having something graphical to indicate the 3D camera centre is quite useful, as it provides expectation of how rotation will work (and also panning in the patch here). I tried decoupling these a while back and reverted it after a month or so as I didn't like the result. If we want to have a 3D cursor for other things perhaps it would be worth trying multiple cursors (in different styles?).

A second set of crosshairs is an absolute must!

> Having two quite different ways to move the centre of the 3D camera seems like a bad thing.

It *is* a bit strange, but I think there is a call for allowing continuous movement. I did receive a lot of informal feedback from people wanting to do something like that for a long time.

I would like to implement continuous movement with WASD keyboard controls and see how that works out. It would be like pushing a wheelbarrow (except you can strafe). Perhaps a third camera mode, "free look" mode, which behaves like a computer game would be intuitive for some users. We would have to absorb the mouse focus (perhaps with double-click to enter and exit free-look?). The crosshairs could be used to select points. The zoom level currently sets the distance scale, so it would have to also set the velocity of movement - otherwise I don't see how we'll get scale-invariant exploration in free look mode.

On the technical side, I'd need to implement an event loop or at least a timer-event whenever a movement key is held.

c42f:
> Probably a good point, but I'd wager there are more mice with a scroll wheel than a good middle button. Touchpads, apple magic mice, etc all support two finger scroll. Does that work currently?

Two finger scroll worked fine on ubuntu when I tried it - it was mapped to scroll events.  Not sure about other OSs, I don't have them sitting around for ease of testing.

I'd definitely welcome a well implemented first-person mode (#15), just not enough to do it myself yet ;-)

c42f:
@andyferris something odd has happened to this branch.  You probably need to rebase against origin/master to make my commits go away.

andyferris:
Right... git and I aren't really friends at the moment. I *did* try to rebase the first commit *on top* of your changes, but I must have done something backwards.

PR #132, opened 2016-09-10
Fix horrible bug: glDeleteBuffers used instead of glDeleteVertexArrays

c42f:
OMSDLFKJG this was hard to find, thank goodness for apitrace.

The bug It leads to some VBOs (particularly that used by PointArray)
being randomly deleted before we're done with them, intermittently
leading to segfaults inside the driver.  Driver- and OS-dependent
madness!

Another effect of this bug was that vertex array handles continually
leaked.  This is probably the cause of some lower powered cards
eventually locking up under resource starvation.

### DIFF
--- a/src/render/Geometry.cpp
+++ b/src/render/Geometry.cpp
@@ -87,7 +87,7 @@ void Geometry::destroyBuffers()
     for (auto& it: m_VAO)
     {
         GLuint vao = it.second;
-        glDeleteBuffers(1, &vao);
+        glDeleteVertexArrays(1, &vao);
     }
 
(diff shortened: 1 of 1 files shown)


andyferris:
I'm so amused that this was a 7 character fix :)

c42f:
Heh, yes.  Probably several hours per character counting all the other things I fixed in pursuit of the root cause here!  Less depressing than a single-character fix I guess :)

andyferris:
Well, it's usually the "needle in a haystack" things that bite worst.

Issue #133, opened 2016-09-12
Modify point attributes in place using index lists


c42f:
For interactive scripts which use displaz as a 3D window, it would be very useful to be able to modify the properties of a subset of vertices in an existing dataset, especially if that dataset is quite large.

An implementation idea is to pass a dataset label, a set of indices and set of attributes to modify, with displaz then making the modifications in place.  The indices should be the indices of vertices loaded from the initial point cloud file (as opposed to the rearranged ordering as stored in memory).

Versioning on the displaz side could be supported if desired, by storing the previous attribute values - this would make implementing undo systems in scripts far easier.

@andyferris - here's an issue to allow modifying point clouds in place.

c42f:
Implementation notes:

Potential command line usage:

```
displaz -modify <my_geom_label> <index_and_attributes>.ply
```

* The "displaz ply" format as it exists is a viable input format for the index list and attributes.
* The FileLoader thread currently only loads instances of `Geometry`, but for modifying an existing Geometry we should produce a `GeometryMutator` or some such thing, which gets applied to an existing geometry on the main thread.
* `-modify` could be `-mutate`, `-update`, ... ?

andyferris:
Seems like a really cool idea Chris; check out a small start at #134.

c42f:
Done in #134

PR #134, opened 2016-09-12
WIP: Mutate a subset of vertices from command line

andyferris:
This is WIP towards an implementation of #133. The idea is to let users change a subset of vertex properties on a subset of verticies, making the IPC much faster for complex and rich scripting applications.

The first commit will allow us to use the original indices to refer to only a subset of points which we wish to modify, but takes slightly more RAM by default. The next steps, as I see them, will be

 * Create a new `GeometryMutator` class that loads a (specially tagged?) .ply file.
 * When the `GeometryMutator` is passed to the main thread, it must search for data of the same label and ask `GeometryMutator` to mutate it.
 * Set up the CLI

### DIFF
--- a/src/fileloader.h
+++ b/src/fileloader.h
@@ -110,12 +110,15 @@ class FileLoader : public QObject
         void loadFileImpl(const FileLoadInfo& loadInfo, bool reloaded)
         {
             // Different codepath for mutating existing data.
+            // TODO Geometry and GeometryMutator have different exception handling
+            //      that could be made more consistent.
             if (loadInfo.mutateExisting)
             {
                 std::shared_ptr<GeometryMutator> mutator(new GeometryMutator());
                 if (!mutator->loadFile(loadInfo.filePath))
                 {
                     g_logger.error("Could not load %s", loadInfo.filePath);
+                    return;
                 }
                 mutator->moveToThread(0);
                 mutator->setLabel(loadInfo.dataSetLabel);
--- a/src/fileloader.h
+++ b/src/fileloader.h
@@ -113,7 +113,10 @@ class FileLoader : public QObject
             if (loadInfo.mutateExisting)
             {
                 std::shared_ptr<GeometryMutator> mutator(new GeometryMutator());
-                mutator->loadFile(loadInfo.filePath);
+                if (!mutator->loadFile(loadInfo.filePath))
+                {
+                    g_logger.error("Could not load %s", loadInfo.filePath);
+                }
                 mutator->moveToThread(0);
                 mutator->setLabel(loadInfo.dataSetLabel);
                 emit geometryMutatorLoaded(mutator);
--- a/src/fileloader.h
+++ b/src/fileloader.h
@@ -115,7 +115,9 @@ class FileLoader : public QObject
                 std::shared_ptr<GeometryMutator> mutator(new GeometryMutator());
                 mutator->loadFile(loadInfo.filePath);
                 mutator->moveToThread(0);
+                mutator->setLabel(loadInfo.dataSetLabel);
                 emit geometryMutatorLoaded(mutator);
+
                 if (loadInfo.deleteAfterLoad)
                 {
                     // Only delete after successful load:  Load errors
--- a/src/fileloader.h
+++ b/src/fileloader.h
@@ -22,15 +22,17 @@ struct FileLoadInfo
     QString dataSetLabel; /// Human readable label for the dataset
     bool replaceLabel;    /// Replace any existing dataset in the UI with the same label
     bool deleteAfterLoad; /// Delete file after load - for use with temporary files.
+    bool mutateExisting;  /// Replace vertex data in-place and discard the result
 
-    FileLoadInfo() : replaceLabel(true), deleteAfterLoad(false) {}
+    FileLoadInfo() : replaceLabel(true), deleteAfterLoad(false), mutateExisting(false) {}
     FileLoadInfo(const QString& filePath_, const QString& dataSetLabel_ = "",
                  bool replaceLabel_ = true)
         : filePath(filePath_),
         dataSetLabel(dataSetLabel_),
         replaceLabel(replaceLabel_),
         // Following must be set explicitly - getting it wrong will delete user data!
-        deleteAfterLoad(false)
+        deleteAfterLoad(false),
+        mutateExisting(false)
     {
         if (dataSetLabel_.isEmpty())
         {
@@ -94,6 +96,8 @@ class FileLoader : public QObject
         void geometryLoaded(std::shared_ptr<Geometry> geom,
                             bool replaceLabel, bool reloaded);
 
+        void geometryMutatorLoaded(std::shared_ptr<GeometryMutator> mutator);
+
     private slots:
         void asyncLoadFile(const FileLoadInfo& loadInfo, bool reloaded)
         {
@@ -105,6 +109,24 @@ class FileLoader : public QObject
 
         void loadFileImpl(const FileLoadInfo& loadInfo, bool reloaded)
         {
+            // Different codepath for mutating existing data.
+            if (loadInfo.mutateExisting)
+            {
+                std::shared_ptr<GeometryMutator> mutator(new GeometryMutator());
+                mutator->loadFile(loadInfo.filePath);
+                mutator->moveToThread(0);
+                emit geometryMutatorLoaded(mutator);
+                if (loadInfo.deleteAfterLoad)
+                {
+                    // Only delete after successful load:  Load errors
+                    // are somewhat likely to be user error trying to load
+                    // something they didn't mean to.
+                    QFile::remove(loadInfo.filePath);
+                }
+                return;
+            }
+
+            // Standard loading code
             std::shared_ptr<Geometry> geom = Geometry::create(loadInfo.filePath);
             geom->setLabel(loadInfo.dataSetLabel);
             connect(geom.get(), SIGNAL(loadProgress(int)),
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -66,6 +66,7 @@ int main(int argc, char *argv[])
 
     bool clearFiles = false;
     bool addFiles = false;
+    bool mutateData = false;
     bool deleteAfterLoad = false;
     bool quitRemote = false;
     bool queryCursor = false;
@@ -108,6 +109,7 @@ int main(int argc, char *argv[])
         "-unload %s",    &unloadFiles,   "Remote: unload loaded files matching the given (unix shell style) pattern",
         "-quit",         &quitRemote,    "Remote: close the existing displaz window",
         "-add",          &addFiles,      "Remote: add files to currently open set, instead of replacing those with duplicate labels",
+        "-modify",       &mutateData,    "Remote: mutate data already loaded with the matching label (requires displaz .ply with an \"index\" field to indicate mutated points)",
         "-rmtemp",       &deleteAfterLoad, "*Delete* files after loading - use with caution to clean up single-use temporary files after loading",
         "-querycursor",  &queryCursor,   "Query 3D cursor location from displaz instance",
         "-script",       &script,        "Script mode: enable several settings which are useful when calling displaz from a script:"
@@ -211,6 +213,11 @@ int main(int argc, char *argv[])
         QByteArray command;
         QDir currentDir = QDir::current();
         command = "OPEN_FILES\n";
+        if (mutateData)
+        {
+            command += QByteArray("MUTATE_EXISTING");
+            command += '\0';
+        }
         if (!addFiles)
         {
             command += QByteArray("REPLACE_LABEL");
--- a/src/ply_io.cpp
+++ b/src/ply_io.cpp
@@ -302,13 +302,10 @@ bool loadPlyVertexProperties(QString fileName, p_ply ply, p_ply_element vertexEl
 // Utilities for loading ply in "displaz-native" format
 
 /// Find all elements with name "vertex_*"
-///
-/// The position element will always be in vertexElements[0] if present.
 bool findVertexElements(std::vector<p_ply_element>& vertexElements,
                         p_ply ply, size_t& npoints)
 {
     int64_t np = -1;
-    int positionIndex = -1;
     for (p_ply_element elem = ply_get_next_element(ply, NULL);
          elem != NULL; elem = ply_get_next_element(ply, elem))
     {
@@ -326,8 +323,6 @@ bool findVertexElements(std::vector<p_ply_element>& vertexElements,
                 return false;
             }
             vertexElements.push_back(elem);
-            if (strcmp(name, "vertex_position") == 0)
-                positionIndex = (int)vertexElements.size()-1;
         }
         else
         {
@@ -335,8 +330,6 @@ bool findVertexElements(std::vector<p_ply_element>& vertexElements,
         }
     }
 
-    if (positionIndex != -1 && positionIndex != 0)
-        std::swap(vertexElements[0], vertexElements[positionIndex]);
     npoints = np;
     return true;
 }
--- a/src/ply_io.cpp
+++ b/src/ply_io.cpp
@@ -334,11 +334,7 @@ bool findVertexElements(std::vector<p_ply_element>& vertexElements,
             g_logger.warning("Ignoring unrecogized ply element: %s", name);
         }
     }
-    /*if (positionIndex == -1)
-    {
-        g_logger.error("%s", "No vertex position found in ply file");
-        return false;
-    }*/
+
     if (positionIndex != -1 && positionIndex != 0)
         std::swap(vertexElements[0], vertexElements[positionIndex]);
     npoints = np;
--- a/src/ply_io.cpp
+++ b/src/ply_io.cpp
@@ -334,12 +334,12 @@ bool findVertexElements(std::vector<p_ply_element>& vertexElements,
             g_logger.warning("Ignoring unrecogized ply element: %s", name);
         }
     }
-    if (positionIndex == -1)
+    /*if (positionIndex == -1)
     {
         g_logger.error("%s", "No vertex position found in ply file");
         return false;
-    }
-    if (positionIndex != 0)
+    }*/
+    if (positionIndex != -1 && positionIndex != 0)
         std::swap(vertexElements[0], vertexElements[positionIndex]);
     npoints = np;
     return true;
--- a/src/render/Geometry.h
+++ b/src/render/Geometry.h
@@ -65,7 +65,10 @@ class Geometry : public QObject
         //--------------------------------------------------
         /// Mutate a geometry
         ///
-        /// Modify some geometry data with the data in a GeometryMutator
+        /// Apply a per-vertex modification of geometry data with the data in a 
+        /// GeometryMutator, which keeps an index to a subset of vertices and
+        /// new data for a subset of fields. The number of vertices remains
+        /// constant.
         virtual void mutate(std::shared_ptr<GeometryMutator> mutator) { }
 
         //--------------------------------------------------
--- a/src/render/Geometry.h
+++ b/src/render/Geometry.h
@@ -13,6 +13,8 @@
 #include <QString>
 #include <QMetaType>
 
+#include "GeometryMutator.h"
+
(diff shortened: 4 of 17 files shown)


c42f:
Yep this makes sense, you're looking in the right place.  One thing I'm struggling with is how to have the memory usage be efficient for large laz files while also have good generality for scripting usage. Having index lists is required for the general case but it really bloats things up (2x 64 bit indices per point for forward and backward indexing adds up to more than the size of the positions.) 

Perhaps the right thing is to store the indices for the general case by default (as you have here), but have a memory efficient mode which strips out a bunch of the less commonly used attributes, and doesn't store the index array.  This would also allow the time attribute to be loaded from laz which would be nice (still would have to figure out how to map the double precision there onto the GPU).

andyferris:
> Perhaps the right thing is to store the indices for the general case by default (as you have here), but have a memory efficient mode

Seems quite reasonable!

andyferris:
Also, do we need to support more than 4 billion points? Couldn't we use `uint_32`, and keep just a single index?

c42f:
In principle I'd like to support more than 2^32 points, though I've never tried to load this much into displaz in practice.  It's a bit of a catch 22 here: 64 vs 32 bits per index only matters if you have a lot of points, which is also the only case where you really care about memory efficiency.

c42f:
For now I'd go with `size_t` for the index, unless you want to opt for a `GeomField` to hold the index itself, in which case you can cater for various integer index types, depending on the total number of points.

andyferris:
>  It's a bit of a catch 22 here: 64 vs 32 bits per index only matters if you have a lot of points, which is also the only case where you really care about memory efficiency.

True!! I'll just use `size_t` for now.

andyferris:
OK there is a compile error I don't understand but otherwise this might be "logically" complete.

c42f:
Looks like the extremely obscure error in the compile

```
error: use of deleted function ‘GeomField::GeomField(const GeomField&)’
```

Is just that you're trying to copy the fields when you return them by value from `GeometryMutator::fields()`, but you can't because `GeomField` is non-copyable as a matter of efficiency.  If you change the signature to

```C++
const std::vector<GeomField>& fields() const { return m_fields; }
```

you can get back to fixing some of the less obscure compile warnings ;-)

andyferris:
lol, right :)

andyferris:
Now its a `QObject` error. Sorry I'm only getting a couple minutes here and there today.

c42f:
Some forgotten headers I think.  You'll need `#include <QObject>` at least.

andyferris:
Now this appears to be working. Needs some more sanity checks before merging:

 * Upon application to a PointArray, check that the indices are within bounds and non-negative (once).
 * If GeometryMutator::loadFile returns false, we should handle that.
 * GeometryMutator::loadFile could do with exception handling.
 * A user-visible warning for when a field is not encountered in the target (but other fields remain applied).

c42f:
Yes, all good things to have :-)

andyferris:
All done. Time for some testing and code review.

andyferris:
As I mentioned, one known caveat is using 32-bit integers for the mutation index. On thing we could do is also maintain the same for the PointArray index, which would have the advantage of halving the memory overhead on 64 bit machines. Estimating a minimum of 16 bytes of vertex data gives over 64GB memory required for 4 billion points (32 bit indices), which won't fit on graphics cards (let alone typical system RAM) for a while yet.

The other option is to hack rply to load 64 bit integers.

andyferris:
I'm having a problem with mutating positions. As far as I can tell, the "position" field is changing but the points are rendered in the same position. @c42f are the vertex positions cached somewhere else?

andyferris:
I think the tree stores indices. Also, I might need to store `offset` as `PointArray::m_offset` in order to do this...

src/fileloader.h
```diff
@@ -105,6 +109,29 @@ class FileLoader : public QObject
 
         void loadFileImpl(const FileLoadInfo& loadInfo, bool reloaded)
         {
+            // Different codepath for mutating existing data.
+            if (loadInfo.mutateExisting)
+            {
+                std::shared_ptr<GeometryMutator> mutator(new GeometryMutator());
+                if (!mutator->loadFile(loadInfo.filePath))
+                {
+                    g_logger.error("Could not load %s", loadInfo.filePath);
```
c42f: Nees a `return;`

c42f: Perhaps just a TODO comment to make exception handling between `Geometry` and `GeometryMutator` consistent.

src/main.cpp
```diff
@@ -108,6 +109,7 @@ int main(int argc, char *argv[])
         "-unload %s",    &unloadFiles,   "Remote: unload loaded files matching the given (unix shell style) pattern",
         "-quit",         &quitRemote,    "Remote: close the existing displaz window",
         "-add",          &addFiles,      "Remote: add files to currently open set, instead of replacing those with duplicate labels",
+        "-modify",       &mutateData,    "Remote: mutate data already loaded with the matching label (requires displaz .ply with an \"index\" field to indicate mutated points)",
```
c42f: So usage is

```
displaz -rmtemp -modify -label mydataset /tmp/blah.ply
```

Seems reasonable, though perhaps not entirely clear usage.  Would `-modifylabel` make more sense as an alternative to `-label`:

```
displaz -rmtemp -modifylabel mydataset /tmp/blah.ply
```

?

src/ply_io.cpp
```diff
@@ -334,12 +334,12 @@ bool findVertexElements(std::vector<p_ply_element>& vertexElements,
...
         }
     }
-    if (positionIndex == -1)
+    /*if (positionIndex == -1)
     {
         g_logger.error("%s", "No vertex position found in ply file");
         return false;
-    }
-    if (positionIndex != 0)
+    }*/
```
c42f: Needs cleanup - need to move the error somewhere sensible I suppose.

src/render/Geometry.h
```diff
@@ -61,6 +63,12 @@ class Geometry : public QObject
         virtual bool loadFile(QString fileName, size_t maxVertexCount) = 0;
 
         //--------------------------------------------------
+        /// Mutate a geometry
+        ///
+        /// Modify some geometry data with the data in a GeometryMutator
```
c42f: Would be good to document the action of the mutator more fully.  Specifically, `GeometryMutator` applies per-vertex changes.

src/render/GeometryMutator.cpp
```diff
@@ -0,0 +1,75 @@
...
+    {
+        std::unique_ptr<t_ply_, int(*)(p_ply)> ply(
+                ply_open(fileName.toUtf8().constData(), logRplyError, 0, NULL), ply_close);
+        if (!ply || !ply_read_header(ply.get()))
+            return false;
+        // Parse out header data
+        p_ply_element vertexElement = findVertexElement(ply.get(), m_npoints);
+        if (vertexElement)
+        {
+            g_logger.error("Expected displaz formated ply for file %s", fileName);
```
c42f: Spelling formated

src/render/GeometryMutator.cpp
```diff
@@ -0,0 +1,75 @@
...
+            m_indexFieldIdx = (int)i;
+            break;
+        }
+    }
+    if (m_indexFieldIdx == -1)
+    {
+        g_logger.error("No index field found in file %s", fileName);
+        return false;
+    }
+    m_index = (uint*)m_fields[m_indexFieldIdx].as<uint>();
```
c42f: A few issues here:
* You shouldn't need the extra cast
* Where does the `uint` type come from (it's not a C type)?  You should use `uint32_t` if that's what you mean.
* What if `spec` doesn't match with `uint`?  That's probably a segfault!

For making typespecs for a scalar type `T`, I tentatively suggest a new metafunction:

```C++
TypeSpec::makeSpec<T>();  // Return correct TypeSpec for unscaled type T
```

src/render/GeometryMutator.h
```diff
@@ -0,0 +1,65 @@
...
+        const QString& label() const { return m_label; }
+
+        /// Set the arbitrary user-defined label for the geometry.
+        void setLabel(const QString& label) { m_label = label; }
+
+        /// Get number of points to mutate
+        size_t pointCount() const { return m_npoints; }
+
+        /// Get index of points to mutate
+        uint* index() const { return m_index; }
```
c42f: `uint`?  I'd prefer `uint32_t` if that's what it is.

src/render/GeometryMutator.h
```diff
@@ -0,0 +1,65 @@
...
+    private:
+
+        /// Label of data to mutate
+        QString m_label;
+        /// Total number of loaded points
+        size_t m_npoints;
+        /// Position offset
+        V3d m_offset;
+        /// Point data field storage
+        std::vector<GeomField> m_fields;
```
c42f: Meta note: We really need a `GeomFields` class to wrap the `std::vector` and offset above.

No need to do this in this PR, just a note on a cleanup I've been thinking about for a while.

src/render/PointArray.cpp
```diff
@@ -368,7 +368,7 @@ bool PointArray::loadFile(QString fileName, size_t maxPointCount)
 
     // Sort points into octree order
     emit loadStepStarted("Sorting points");
-    std::unique_ptr<size_t[]> inds(new size_t[m_npoints]);
+    std::unique_ptr<size_t[]> inds = std::unique_ptr<size_t[]>(new size_t[m_npoints]);
```
c42f: Code churn?  Any reason this was changed?

src/render/PointArray.cpp
```diff
@@ -386,14 +386,83 @@ bool PointArray::loadFile(QString fileName, size_t maxPointCount)
     {
         g_logger.debug("Reordering field %d: %s", i, m_fields[i]);
         reorder(m_fields[i], inds.get(), m_npoints);
-        emit loadProgress(int(100*(i+1)/m_fields.size()));
+        emit loadProgress(int(100*(i+1)/(m_fields.size()+1))); // denominator +1 for permutation reorder below
     }
     m_P = (V3f*)m_fields[m_positionFieldIdx].as<float>();
+    // The index we want to store is the reverse permutation of the index above
+    m_inds = std::unique_ptr<size_t[]>(new size_t[m_npoints]);
```
c42f: Should we make this `uint32_t`, given that you're assuming that elsewhere anyway?

src/render/PointArray.cpp
```diff
@@ -386,14 +386,83 @@ bool PointArray::loadFile(QString fileName, size_t maxPointCount)
     {
         g_logger.debug("Reordering field %d: %s", i, m_fields[i]);
         reorder(m_fields[i], inds.get(), m_npoints);
-        emit loadProgress(int(100*(i+1)/m_fields.size()));
+        emit loadProgress(int(100*(i+1)/(m_fields.size()+1))); // denominator +1 for permutation reorder below
     }
     m_P = (V3f*)m_fields[m_positionFieldIdx].as<float>();
+    // The index we want to store is the reverse permutation of the index above
```
c42f: Let's have a note about why we're storing the indexing order.

src/render/PointArray.cpp
```diff
@@ -386,14 +386,83 @@ bool PointArray::loadFile(QString fileName, size_t maxPointCount)
...
 
 
+void PointArray::mutate(std::shared_ptr<GeometryMutator> mutator)
+{
+    // Now we need to find the matching columns
+    auto npoints = mutator->pointCount();
+    const std::vector<GeomField>& mut_fields = mutator->fields();
+    auto mut_idx = mutator->index();
+
+    if (m_npoints > 4294967296)
```
c42f: Magic constant should be replaced with `UINT32_MAX`

src/render/PointArray.cpp
```diff
@@ -386,14 +386,83 @@ bool PointArray::loadFile(QString fileName, size_t maxPointCount)
...
+        g_logger.error("Mutation with more than 2^32 points is not supported");
+        return;
+    }
+
+    // Check index is valid
+    for (size_t j = 0; j < npoints; ++j)
+    {
+        if (mut_idx[j] < 0 || mut_idx[j] >= m_npoints)
+        {
+            g_logger.error("Index out of bounds - got %d (should be between zero and %d)", mut_idx[j], m_npoints-1);
```
c42f: Yep, nice.

src/render/PointArray.cpp
```diff
@@ -386,14 +386,83 @@ bool PointArray::loadFile(QString fileName, size_t maxPointCount)
...
+            return;
+        }
+    }
+
+    // Currently doesn't matter if a field isn't found
+    // Convenient way to ignore "index" but is not robust
```
c42f: Comment about "index" out of date (you have a continue below)?

src/render/PointArray.cpp
```diff
@@ -386,14 +386,83 @@ bool PointArray::loadFile(QString fileName, size_t maxPointCount)
...
+    {
+        if (mut_fields[mut_field_idx].name == "index")
+            continue;
+
+        bool found = false;
+
+        for (size_t field_idx = 0; field_idx < m_fields.size(); ++field_idx)
+        {
+
+            if (m_fields[field_idx].name == mut_fields[mut_field_idx].name)
```
c42f: Perhaps just for the sake of indentation and code flow, it might be nicer to make this loop break when you find the right name, and set an index instead of setting `found`.  (This would also be better if we had `GeomFields::findField()` or some such thing.)

Then do the actual modification with the memcpys out of the field finding loop.

src/typespec.h
```diff
@@ -81,6 +81,13 @@ struct TypeSpec
 
     /// Get number of bytes required to store the field for a single point
     int size() const { return elsize*count; }
+
+    bool operator==(const TypeSpec& other) const
```
c42f: One line of docs would be worth it.  Perhaps "Check that two TypeSpecs are exactly equal"

c42f:
Aww, I'm not so keen on the big red cross that the new github UI puts on the PR when you request changes.  It doesn't reflect that I'm super happy to have this ;-)

c42f:
> @c42f are the vertex positions cached somewhere else?

Nope, shouldn't be.  I don't understand this.

src/render/PointArray.cpp
andyferris: It would be cool if programming languages supported `break(N)` and `continue(N)` since these can be implemented with a static `goto`...

src/render/PointArray.cpp
andyferris: Just in case it changes one day? ;)

Will do. The thing about magic macros is that you have to know they exist. :)

src/render/PointArray.cpp
andyferris: I'm ok with this. This might limit us to 4 billion points even when not mutating...

src/render/PointArray.cpp
andyferris: It *was* churned (deleted and re-added later when I changed what I was doing). Consider it unchurned.

src/render/GeometryMutator.cpp
andyferris: Now fixed on line 58. I skipped the metafunction for now and implemented `TypeSpec::uint32_i()`. But yes a metafunction would be awesome.

src/main.cpp
andyferris: It might, if you think so, but I kind-of liked the way the interface didn't really change - there is only one way to specify a label.

src/ply_io.cpp
andyferris: Yes, damn I forgot about that one...

src/typespec.h
andyferris: So the "docs" are all the things with with three slashes?

src/ply_io.cpp
andyferris: Actually isn't this already covered? It will be caught in `PointArray::loadFile`, for instance. Where do you expect it to go wrong?

andyferris:
Thanks Chris. I've done a round of changes - but I still am not sure if that error I removed from ply_io.cpp was necessary or not.

andyferris:
Hmm... something just broke because my app stopped updating colors...

src/typespec.h
c42f: Yep.  This is just the doxygen convention.  I'm not actually using doxygen, but I like the conceptual separation of three slashes for "public API" documentation from two slashes for implementation detail comments.

src/render/PointArray.cpp
c42f: For breaking multiple loops?

Would occasionally be helpful, though I usually find it's a sign that a helper function is in order (`return` can be used just fine from deeply nested loops)

src/render/PointArray.cpp
c42f: Sure no probs, thanks.

src/main.cpp
c42f: There's not a huge amount in it... and it's going to be used almost entirely from scripts anyway, so let's go with what you have.

src/render/PointArray.cpp
```diff
@@ -386,14 +386,85 @@ bool PointArray::loadFile(QString fileName, size_t maxPointCount)
...
     return true;
 }
 
 
+void PointArray::mutate(std::shared_ptr<GeometryMutator> mutator)
+{
+    // Now we need to find the matching columns
+    auto npoints = mutator->pointCount();
+    const std::vector<GeomField>& mut_fields = mutator->fields();
+    auto mut_idx = mutator->index();
```
c42f: I only just noticed - all the variable names in this function are using `underscore_case` which is a bit inconsistent.

c42f:
I pushed a change to fix the "position" field by applying the offsets as required.

When testing, I notice that getting the field type correct can be difficult - probably not bad if you do it all from the scripting layer, but annoying when a file is loaded and then you subsequently try to modify a part of that from the script.

andyferris:
How does the offset thing work? Does PointArray remember the offset it first used so the user's mutator doesn't need to specify it explicitly?

c42f:
Yes, exactly.  The offset is remembered as a double vector on the CPU side, and composed with the camera transformation on the CPU side which is also stored in double precision.

andyferris:
Oh? I figured the offset allowed us to do everything graphical in Float32 and leave all user-facing positions in UTM or whatever, and have the camera transformation on the GPU. 

So all rendered vertices are sent to the GPU every frame, in camera coordinates? I guess otherwise it would make it harder to implement your partial rendering to maintain frame rate (unless you used CUDA or something, maybe a *Compose.jl*-style approach to the OctTree for culling... I dunno).

c42f:
No no, nothing so crazy ;-)  All model->camera transformations are done in the vertex shader as usual, with the sole exception of removing a large mostly-shared offset from the data and camera.

Suppose you're in ECEF coords, you'll have typical data coordinate values ~ the radius of the earth away from the origin which is ~6e6 m.  But `eps(Float32(6e6)) = 0.5`, so the available precision in float is only half a meter.  But using double precision on GPUs is generally not great.  Displaz stores positions as float, but relative to a double origin `offset`.  The true coordinates of point `i` are then `offset + position[i]`.

On the CPU side, the camera has `origin_camera` also in double precision; the point is that `offset - origin_camera` is generally rather small and can adequately be represented in float precision - this is then what's sent to the GPU.

c42f:
To be clear, `offset` is stored per-geometry (just using the position of the first vertex for simplicity), and this works well, provided vertices in a given geometry all have a similar vector from the origin.

andyferris:
Right, OK, that is more like what I first expected. :)

src/ply_io.cpp
c42f: At render/PointArray.cpp:265

c42f: It's a bit of a mess in there, I'll make a commit to sort it out.

c42f: Should be ok now I think.

src/render/PointArray.cpp
c42f: Renamed.

c42f:
I messed around with a few more things, I think this is probably ready to merge.

andyferris:
Looks great to me. All the offsets etc make sense to me (though I'm still not convinced that's a ply loader's job to compute rather than a Geometry), and the other changes are good. Thanks, Chris!

PR #142, opened 2016-10-01
Resolve Coverity warning - dynamic_cast can return NULL, add check

nigels-com:
dynamic_cast


c42f:
Thanks for these fixes, I've merged them all in.

If a single PR would have been easier for you with the coverty fixes, I'd also have been fine with that as it's all small nonintersecting stuff.

PR #145, opened 2016-10-01
Update GLEW to release 2.0.0

nigels-com:
GLEW 2.0.0 was released in July 2016.


c42f:
Thanks, I think this will let me remove the `glewExperimental` hack that @r-chris had to put in when porting to OpenGL 3 :-)

c42f:
Rebased & merged, thanks.

Presumably 69e5a60cce2e267ff109f0043e87e40a7338efa0 is now safe?  (Seems to work)

Issue #136, opened 2016-09-16
Idea: Scripting access to logger


andyferris:
At the moment there are few options that a scripting/hooks application can use to interact with displaz's user. One simple way would be to allow a scripting command for adding a log entry, like:

```
displaz -loginfo "Welcome to my app"
displaz -logwarning "You did the wrong thing"
displaz -logerror "You REALLY did the wrong thing"
```

These commands would be able to be used with or without a command to load a file or mutate some data. For completeness, to match the Qt logger, we could also implement `-logdebug`, `-logcritical` and `-logfatal` (though I'm not sure whether or not the last should apply if it's just the scripting app that dies).

c42f:
Yes, this would be good.  I'd probably go for a single command line option, perhaps

```
displaz -notify <level> <message>
```

would be clearer naming?   `displaz -notify error "You are a bad, bad user."`

andyferris:
+1

Perhaps in the future there may be more ways to interact with the user, e.g. through the log or through a pop-up dialogue or a widget in the OpenGL window or whatever. I'm wondering if "notify" the correct word to indicate specifically the log?

andyferris:
I guess these could be 

```
displaz -notify popup "You made me angry, and now I will make you angry with this pop-up notification"
```

or whatever

c42f:
lol.

Yes, having a string as the notification type is pretty extensible (could have a simple format `$style:$severity`, for example: `popup:error` `log:error` `status:info`).  Though a popup dialog style is unlikely to ever be merged :-P

c42f:
Closed by #148

PR #148, opened 2016-10-14
Remote command -notify for scripting access to log window

c42f:
Implement only log:<level> for the logspec, for now.  Potentially more
in the future.  @andyferris how does this look for #136 ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,10 +28,19 @@ before_install:
     - if [ "$TRAVIS_OS_NAME" == "osx" ]; then brew update; fi
 
 install:
-    - if [ "$TRAVIS_OS_NAME" == "linux" ]; then sudo apt-get install -qq -y git g++ cmake qt5-default python-docutils; fi
-    - if [ "$TRAVIS_OS_NAME" == "osx" ]; then brew install qt5; fi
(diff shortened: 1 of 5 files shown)


andyferris:
Beautiful. :) Thanks Chris

c42f:
Cool, I'll merge this then, and we can see how it goes.

c42f:
Huh, hold on, travis is failing.

c42f:
Yikes, that's quite a bug at main.cpp:129 caught by travis.  I have *no* idea why this worked locally in testing, it should have segfaulted nicely, or something.

c42f:
Well, the reason that worked at all was super obscure:

https://gcc.gnu.org/onlinedocs/gcc/Conditionally-supported-behavior.html
http://stackoverflow.com/questions/10083844/passing-non-pod-type-to-variadic-function-is-undefined-behavior

So gcc now allows classes to be passed to C-style variadic functions, and silently converts them to references which happened by ABI chance to be compatible with pointers on the other side (I guess).  Ugh.

andyferris:
Right. Fascinating!

c42f:
Ah, seems the OSX build failures were due to Travis OSX woes - https://www.traviscistatus.com/

PR #149, opened 2016-10-21
Tolerate 2D PLY with missing z property, assumed to be all zero

nigels-com:


nigels-com:
ping.

src/render/TriMesh.cpp
```diff
@@ -90,14 +92,24 @@ static int vertex_cb(p_ply_argument argument)
...
-    info.verts.push_back((float)v);
+    // Append x, y or z
+    if (info.gotZ || (i < 2))
+    {
+        info.verts.push_back(float(v-info.offset[i]));
+    }
+    // Default z to zero as necessary
+    if (!info.gotZ && (i == 1))
+    {
+        info.verts.push_back(-info.offset[i]);
```
c42f: Technically I think this should be `-offset[2]`, though as it stands that will be zero anyway.  Ah, also the reason I had that annoying looking `float` cast in there was to silence warnings on MSVC, I think.

c42f: Heh, actually this is a bug, but the test case has first vertex at (0,0) so it's invisible.

c42f: How about a25a356 ? I also changed the order of verts in the test so this bug would be visible.

nigels-com: Yep, looks good to me.  Thanks for the second opinion.

Issue #152, opened 2016-11-10
IPC : client not getting response


glandais:
Hi,
On a Windows 7 64bits, -querypoint and -hook are not always working on my computer.
When I run "displaz -querycursor", the message is sent to the GUI but response looks blocked somewhere. When the client times out, PointViewerMainWindow::handleMessage is executed.

In main.cpp, if I print on stdout before and after sending the message, it gets the response...
I tried a milliSleep but it didn't worked...

I'm available for any idea/bugfix!

c42f:
Ah, more IPC woes! It's surprisingly hard to make this work nicely on all platforms.  I've had problems with using qt sockets in synchronous mode before.  I worked around a related issue in 8364c38 (see also https://bugreports.qt.io/browse/QTBUG-47683, which has *still* not been addressed...), so the commentary there might supply a place to start looking.

In fact, looking through the code, I suspect the problem may be in `IpcChannel::receiveMessage()`, which calls `waitForReadRead()`, which IIRC is one of the *synchronous* `QLocalSocket` methods, and is probably broken on windows for the same reason that `waitForDisconnected()` is broken. I suggest that you have a read of the commentary in `IpcChannel::waitForDisconnected` and see whether you can run a similar local event loop as a workaround.

A further trawl through the code shows another place which a `waitFor` style synchronous method is called in `IpcChannel::connectToServer()` - this could also be a source of trouble given the general brokenness of this stuff on windows.

As a more general fix for the unreliability of the sychronous API, perhaps we could consider refactoring `IpcChannel` so that it always runs an event loop internally and uses the async functionality under the hood to simulate synchronous behaviour.

c42f:
If you're feeling particularly energetic, another IPC related problem is `InterProcessLock_test` (run with `ctest -R InterProcessLock_test`) which seems broken on windows for reasons I haven't had time to investigate.  I doubt this is the source of your current problem, however.

glandais:
It would have been great to refactor the event loop but it is a bit too hard for me...
Sadly I don't have enough space disk for Visual Studio on my C:, so I have to use Appveyor to build displaz... Then I'm also not able to fix the InterProcessLock_test...

glandais:
I reverted the modifications on `IpcChannel::connectToServer()`, as it locks the program exit when I'm using `displaz -script`...

cbecker:
[...]

PR #153, opened 2016-11-14
Fixing issue #152

glandais:
Fixing #152 
Tested on Windows 7 64bits


c42f:
Sorry I've taken so long to get back to this.  I've just spent the last week generally improving displaz on windows (fixing several packaging problems, and the annoying console window which pops up in the background).

I've done a little cleanup on your branch, as the `-hook` command wasn't disconnecting when the GUI shut down.  I also removed the commented out parts for now (I'm not sure how `waitForConnected()` works on windows, but we don't seem to have problems there in contrast to the other `wait*()` functions)).

PR #155, opened 2016-11-15
Replace .has_key for python bindings

andrewdalecramer:
dict.has_key is only available in python2.x, the `key in dict` syntax is
available in python2.3 and later. This should allow the python bindings
to work for more versions.


c42f:
Looks good, thanks!

Issue #158, opened 2016-11-29
Build fails on a fresh install of Ubuntu 16.10


mcjohnalds:
```
cd build
cmake ..
make -j4
```

Fails with
```
...
[ 91%] Built target displaz-gui
CMakeFiles/unit_tests.dir/typespec.cpp.o:(.qtversion[qt_version_tag]+0x0): undefined reference to `qt_version_tag'
collect2: error: ld returned 1 exit status
...
```

nigels-com:
What's the version of cmake, and Qt?

c42f:
I've observed this in a couple of other machines now. In those cases it appeared when a version of conda shipped qt binaries into the LD_LIBRARY_PATH which were different from the ones which cmake found the headers for.

In that case, unloading conda from the environment fixed that.  Could this be something similar?

mcjohnalds:
I can't find anything related to conda on my system, nor have I ever installed it. It's a fresh install of 16.10 with only git, google-chrome, and the atom code editor installed.

As in my patch adding -DQT_NO_VERSION_TAGGING fixes things.

nigels-com:
Thanks for the information.  Seems like as good excuse as any to spin up a Yakkety docker container, to troubleshoot further.

c42f:
Exactly my thoughts, which is in fact what I just did.

The underlying problem seems to be that some Qt headers are coming in via an include of glutil.h into typespec.cpp, and the very action of transitively including qversiontagging.h causes a linker error, regardless of even calling any Qt functions.

Two obvious options present themselves: (1) just link `util_tests` against Qt::Core, even though it shouldn't be required (2) Disentangle typespec.cpp from the Qt headers entirely.

nigels-com:
For what it's worth, I don't see the harm in linking `util_test` against Qt::Core.

c42f:
Fixed in #161

PR #159, opened 2016-11-29
Fixed "undefined reference to qt_version_tag" on Ubuntu 16.10

mcjohnalds:
Fixes issue #158.

I'm not sure what side effects this switch could have when building on other linux distros.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -186,9 +186,9 @@ else()
         set_target_properties(displaz-gui PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
     endif()
     if (UNIX)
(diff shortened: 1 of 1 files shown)


c42f:
Perhaps we're suffering from the "ugly hack warning" in the qt source :-)
https://github.com/qt/qtbase/blob/1dcd61d553e25c8986e7e6e65efacc51c3cf3bee/src/corelib/global/qversiontagging.h#L49

At a quick look, I'm afraid I don't fully understand what's going on there, and what the wider implications are of turning it off.

Do you have anything unusual in your environment, or is it a truly clean ubuntu 16.10 install?

c42f:
Thanks for the workaround!  Let's do this via #161 instead as it addresses the underlying problem.

Issue #147, opened 2016-10-12
Text annotations in 3D window


c42f:
To annotate objects via the scripting bindings, it would be really useful to have arbitrary billboarded text in the 3D window.

The simplest way to communicate this from the scripting bindings would probably be to have text passed directly from the command line, eg, `displaz -text $x $y $z "Some text"`.

CC @Ross-Batten

c42f:
Valve's signed distance field technique for font rendering seem to be the current sweet spot of good looking and practical for text rendering. It would be a lot of work, but luckily the qt libraries already have some of it done.  Unfortunately it's part of QtQuick 2, so it might be hard to integrate
https://www.kdab.com/integrating-opengl-with-qt-quick-2-applications-part-1/

nigels-com:
If we can live without zooming, the conventional freetype-based approach ought to suffice.
Such as: https://github.com/rougier/freetype-gl

c42f:
Thanks for that link, it seems that freetype-gl is not ftgl (a much older library), and does in fact support the signed distance formulation.  Unclear whether it'll work on windows and OSX though.

The annoying thing about library choice here is that I really don't want freetype and potentially a  bunch of other deps (harfbuzz?) when I've already got Qt for supposedly handling that stuff.  But as usual, if you don't want to do things "the qt way", then you're mostly out of luck :-(  So I can take on some extra deps, write a bunch of code, or try to shoehorn qt into doing what I want.

nigels-com:
Yeah, I never considered Qt first-class for OpenGL stuff.  _Give me a context and get out of my way_
But good enough nowadays?  Perhaps.

mcjohnalds:
QPainter can draw text straight onto a QImage. If that works out well enough I'll have a patch ready soon.

mcjohnalds:
Should the text shrink when you're further away or stay at a fixed size regardless of distance?

c42f:
Both of these seem like valid options in different circumstances.  To choose only one of them, I'd probably go with a fixed size for now, treating the text as data labels.

To generalize a bit, perhaps we could have a configurable minimum and maximum distance: closer than the minimum distance, the text has a fixed maximum size.  Further than the maximum distance it shrinks via the usual geometric scaling.  In between, it's a fixed size.

I'm not sure this makes sense, testing it out in a prototype would help.

c42f:
Closed by #160

PR #160, opened 2016-12-03
Added billboard text annotations

mcjohnalds:
Fixes issue #147 

Example:
```
displaz autzen.laz &
displaz -annotate "exploded thing" 637156.720 850102.120 485.140 &
displaz -annotate "garbage" 636837.230 851533.400 428.570 &
displaz -annotate "delicious eggplant 🍆" 638043.230 851172.570 421.130
```
Result:
annotations

- Annotations stay at a fixed size regardless of distance
- Annotations always face the camera
- Annotations are not obscured by scene geometry
- You can toggle displaying annotations by clicking View->Draw Annotations.

I got the automatic switching between fixed-size and geometric scaling based on distance from camera working (as described in the comments) but I couldn't work out the maths required for a smooth transition between the two so I left that out for now. I'll try and get that working later.

### DIFF
--- a/shaders/annotation.glsl
+++ b/shaders/annotation.glsl
@@ -2,14 +2,14 @@
 // Copyright 2015, Christopher J. Foster and the other displaz contributors.
 // Use of this code is governed by the BSD-style license found in LICENSE.txt
 
-// Draws billboards, which are textured quads that always face the camera,
+// Draws annotations, which are textured quads that always face the camera,
 // arent't obscured by other objects, and don't shrink when they're further
 // away.
 
 uniform mat4 modelViewMatrix;
 uniform mat4 projectionMatrix;
 uniform mat4 modelViewProjectionMatrix;
-uniform vec2 billboardSize; // In pixels
+uniform vec2 annotationSize; // In pixels
 uniform vec2 viewportSize; // In pixels
 
 //------------------------------------------------------------------------------
@@ -24,7 +24,7 @@ void main()
 {
     gl_Position = modelViewProjectionMatrix * vec4(0, 0, 0, 1.0);
     gl_Position /= gl_Position.w;
-    gl_Position.xy += position * billboardSize / viewportSize * 2;
+    gl_Position.xy += position * annotationSize / viewportSize * 2;
     ftexCoord = texCoord;
 }
 
--- a/shaders/annotation.glsl
+++ b/shaders/annotation.glsl
@@ -15,6 +15,8 @@ uniform vec2 viewportSize; // In pixels
 //------------------------------------------------------------------------------
 #if defined(VERTEX_SHADER)
 
+// Position of the current quad vertex. The bottom-left vertex is at (-1,-1)
+// and the top-right at (1,1).
 in vec2 position;
 in vec2 texCoord;
 
@@ -24,7 +26,7 @@ void main()
 {
     gl_Position = modelViewProjectionMatrix * vec4(0, 0, 0, 1.0);
     gl_Position /= gl_Position.w;
-    gl_Position.xy += position * annotationSize / viewportSize * 2;
+    gl_Position.xy += position * annotationSize / viewportSize;
     ftexCoord = texCoord;
 }
 
--- a/shaders/billboard.glsl
+++ b/shaders/billboard.glsl
@@ -1,0 +1,47 @@
+#version 150
+// Copyright 2015, Christopher J. Foster and the other displaz contributors.
+// Use of this code is governed by the BSD-style license found in LICENSE.txt
+
+// Draws billboards, which are textured quads that always face the camera,
+// arent't obscured by other objects, and don't shrink when they're further
+// away.
+
+uniform mat4 modelViewMatrix;
+uniform mat4 projectionMatrix;
+uniform mat4 modelViewProjectionMatrix;
+uniform vec2 billboardSize; // In pixels
+uniform vec2 viewportSize; // In pixels
+
+//------------------------------------------------------------------------------
+#if defined(VERTEX_SHADER)
+
+in vec2 position;
+in vec2 texCoord;
+
+out vec2 ftexCoord;
+
+void main()
+{
+    gl_Position = modelViewProjectionMatrix * vec4(0, 0, 0, 1.0);
+    gl_Position /= gl_Position.w;
+    gl_Position.xy += position * billboardSize / viewportSize * 2;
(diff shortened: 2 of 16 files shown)


shaders/billboard.glsl
```diff
@@ -0,0 +1,47 @@
+#version 150
+// Copyright 2015, Christopher J. Foster and the other displaz contributors.
+// Use of this code is governed by the BSD-style license found in LICENSE.txt
+
+// Draws billboards, which are textured quads that always face the camera,
```
nigels-com: I don't think these are billboards in the usual usage of the term.  I suggest renaming the shader `annotations.glsl`

src/render/Billboard.cpp
```diff
@@ -0,0 +1,122 @@
+// Copyright 2015, Christopher J. Foster and the other displaz contributors.
+// Use of this code is governed by the BSD-style license found in LICENSE.txt
+
+#include "Billboard.h"
```
nigels-com: SImilarly for `Billboard.cpp` and `Billboard.h`, rename to `Annotation.h/cpp`

nigels-com:
Looks good to me.  I'd suggest replacing the billboarding terminology with annotation.  Billboarding is usually for approximating 3D geometry with pre-rendered textures.

mcjohnalds:
Everything's been renamed from billboard to annotation.

shaders/annotation.glsl
```diff
@@ -0,0 +1,47 @@
...
+uniform mat4 modelViewMatrix;
+uniform mat4 projectionMatrix;
+uniform mat4 modelViewProjectionMatrix;
+uniform vec2 annotationSize; // In pixels
+uniform vec2 viewportSize; // In pixels
+
+//------------------------------------------------------------------------------
+#if defined(VERTEX_SHADER)
+
+in vec2 position;
```
c42f: A couple of comments about the coordinate system of `position` would be helpful here or down below: `makeVAO`, `Annotation::draw()` and the shader work together in a tightly coupled way, and it's just a little tricky to figure out what's going on.

mcjohnalds:
Not sure why the dates on my previous commits got garbled when I pushed but there we go.

c42f:
I think this is almost ready to merge; as a first go at an annotation system it's already going to be very useful.  I noticed you're missing one small thing which is to call `restartRender()` inside `addAnnotation()` to force the labels to show up immediately.  (displaz has an event-based OpenGL loop, and doesn't draw anything when the view isn't changing.)

Also, I'd like to explore the command line API just a little more so we can have something forward compatible for scripts to rely on.  Eg, what if we later want a feature to allow people to be replace existing annotations on a geometry?  It would be nice if `-annotate` can expand to cover that use case (and potentially others).

I don't know what's nicest here, just throwing out a few ideas:

```
displaz -annotation "text=$text,parent_geometry=$label" $x $y $z

# Hmm, how does this work out if we wanted other annotations, eg an arrow marker?
displaz -annotation "style=arrow,parent_geometry=$label" $x $y $z

# Or if we wanted to remove a text annotation
displaz -annotation "remove=true,text=$text,parent_geometry=$label" $x $y $z
```
Hmm, are x,y,z are irrelevant in the case of removal?  How should the user refer to the identity of an annotation?  An optional label for an annotation (distinct from the text) might make sense, in analogy to the geometry labelling system which I think is working quite well.

Thoughts?

mcjohnalds:
If you anticipate displaz's API growing a lot it might be worth doing what git does with its subcommands e.g,

```
displaz add [-label <label>] <path>
displaz unload <label>
displaz annotation [-label <label>] [-arrow] <text> <x> <y> <z>
...
```

That way you can overload commands like annotation with lots of options while keeping a straightforward interface. You could leave the existing interface deprecated and undocumented. Ignoring that, if we were to give annotations their own labels, then `-label <label>` and `-unload <label>` could be made to work with annotations, removing the need for parent geometry. E.g,

```
displaz -annotation "a string" 12.3 45.6 78.9 # "annotation1" label generated
displaz -label "foo" -annotation "another string" 98.7 65.4 32.1 # "foo" label generated
displaz -unload "annotation1" # first annotation removed
```

I'll wait until we finalise the API design before I mess with the signals and slots code to call `restartRenderer()`. In fact you might want to leave this PR on the backburner until the API is figured out.

Side note: If you want a simple arrow to point to things then using any of 🡳🡻🢃▼🠷 works fine.

c42f:
Good points, I like the look of the subcommands version.

The displaz command line has slowly grown and at this stage it's definitely getting a bit inconsistent and unwieldy.  My rough plan is to keep growing it organically a while longer and then revisit it in a big refactor.  For the moment I'm putting off the refactor because I want to have a bigger list of operations before trying to generalize to something nicer.  I do wonder whether the definitive API should be in a documented socket protocol (rather than the mess I have now :-)), with the command line just a convenient wrapper around that.

Anyway, that's future stuff.  For the moment, `-label` will do nicely for labelling annotations, but I do think attaching them to parent geometry would be a very useful feature.  (Related - one thing I've been considering for a long time is having geometry lifetime and grouping managed in a tree structure, but I just haven't had the time to do this properly.)

Perhaps the right thing is to just let users rely on naming conventions for now, rather than having the ownership semantics of annotations vs geometry being strangely inconsistent.  This would be quite usable from a script:

```
displaz -label blah some/geometry.ply
displaz -label "blah:foo" -annotation "foo" 1 2 3 # pure label naming convention to indicate grouping
displaz -label "blah:foo2" -annotation "foo2" 4 5 6
displaz -label "blah:foo" -annotation "foo3" 7 8 9  # Replaces "blah:foo"

displaz -unload "blah*" # remove blah and its annotations
```

Then later on, if tree-based grouping of geometries is added, annotations could just be part of that.

mcjohnalds:
To clarify, if I did

```
displaz -label blah some/geometry.ply
displaz -label "blah:foo" -annotation "foo" 1 2 3
displaz -unload blah
```

then the "blah:foo" annotation should still be there?

c42f:
Cprrect.  `-unload` is a bit of a special case in that it takes a shell glob-style pattern.  So `blah*` matches `blah:foo`, but `blah` does not, according to the usual shell globbing rules.

mcjohnalds:
-annotation should now work as you've described

c42f:
This is a great addition, and I'm really sorry to have taken so long to get back to it.  I'm just going to merge this now!

PR #161, opened 2016-12-06
Fix undefined ref to qt_version_tag on ubuntu 16.10

c42f:
The action of transitively including qversiontagging.h causes a linker
error in ubuntu 16.10, regardless of ever calling any Qt functions -
link against QtCore to prevent this.


c42f:
@mcjohnalds I think this is a more root-cause fix for #158.  Does it work for you?

mcjohnalds:
Works 100%. Much better than a workaround.

c42f:
Great :-) I'll wait for the travis build to complete and merge it.

Issue #89, opened 2016-01-30
Reduce required GLSL version where possible


c42f:
It seems that the update to OpenGL 3.2 core for OSX has unnecessarily broken usage on some older GPUs, which can be made to work again by downgrading the required GLSL `#version`.

Unfortunately it may be necessary to set the version programmatically depending on the system.  @chrisidefix - any thoughts? I imagine you had a good reason to specify `#version 150`.

nigels-com:
I would guess that `#version 150` was chosen to match OpenGL 3.2.  I'd guess that `#version 130` (OpenGL 3.0) would suffice in most cases.  (Worth testing the shaders with `130`, probably most of them will be fine.)

c42f:
Yes, that sounds right, IIRC it was something about the OSX driver rejecting earlier versions, but I'll have to hear from @chrisidefix to know for sure.  I did downgrade to `140` for Omri's laptop and things seemed to work fine.

chrisidefix:
Yes, I'm fairly sure `#version 150` was a mandatory change to get the shaders loaded under OS X. I guess we gave to go with a platform specific solution on this one.

c42f:
Thanks.  Bit of a bummer, but should be easy if not pretty.

malthe:
Would it make sense to support `#version 120` – ?

The Intel HD Graphics adapter in my laptop complains that it supports only GLSL 1.00, 1.10 and 1.20.

c42f:
@malthe ideally the shaders would just declare the version they actually need - if you change the `#version` headers to 1.2 do they just work on your laptop?  If not, they'll also need more hacks to avoid relying on newer features not supported on Intel.

For OSX we could have the system munge the shaders to replace any declared version with `#version 1.5`.  It's not pretty but might work out.

malthe:
The GLSL files are written for version 1.5 and they do rely on newer features such as ``flat in``. But it would be cool to have a fallback set of shaders that work on 1.2 :-)

c42f:
Sure, I'm happy to have some backward compatibility and I'd welcome any help in making things work more generally.  One of my main problems is a rather limited set of OS and hardware to test on, so with OpenGL drivers being so variable it's hard to keep things working reasonably on various systems.  I'm not sure there's any alternative other than buying a selection of older hardware and testing on each one.  @nigels-com you've got some experience there, is that about right?

[...]

PR #162, opened 2017-01-12
Include sys/stat.h to fix build on OpenBSD

landryb:
I've been able to build displaz on OpenBSD, using a bit of cmake magic to find the GL headers in usr/X11R6/include - and all i needed was this oneliner.

sadly, the log window says at startup:
```
OpenGL implementation:
  GL_VENDOR    = X.Org
  GL_RENDERER  = Gallium 0.4 on AMD RV610 (DRM 2.29.0 / 6.0)
  GL_VERSION   = 3.1 (Core Profile) Mesa 13.0.2
  GLSL_VERSION = 1.40
  GLEW_VERSION = 2.0.0
ERROR: Could not read cursor shader.
ERROR: Could not read axes shaders
ERROR: Could not read grid shader.
ERROR: Couldn't open shader file "shaders:las_points.glsl": No such file or directory
```
I can open a las file, but it wont display:
```
Loaded 1018103 of 1018103 points from file /tmp/10651_CRAIG_SemisLidar_L93-IGN69_694500_6511250.las in 0.81 seconds
```
What could be missing ?

### DIFF
--- a/src/InterProcessLock.cpp
+++ b/src/InterProcessLock.cpp
@@ -13,6 +13,7 @@
 #else
 #   include <unistd.h>
 #   include <sys/file.h> // flock()
+#   include <sys/stat.h> // S_IWUSR defines
 #   include <string.h>   // strerror()
 #endif
 


landryb:
Ah, making some progress, it was complaining about the missing shaders because i didnt *install* displaz to somedir and was trying to run it from the src dir. If i manually open the shader file via the shader menu, it complains about the GLSL version:

```
ERROR: Could not compile shader:
0:1(10): error: GLSL 1.50 is not supported. Supported versions are: 1.10, 1.20, 1.30, 1.40, and 1.00 ES

```
i suppose the GLSL version i have is too old then ?

landryb:
And finally managed to properly display a las file, patching all the glsl files to with 'version 140' instead of 'version 150' as header. I'll blame our too old mesa. displaz looks awesome :)

landryb:
Correction, our mesa isnt too old, it's my gfx card which doesnt support GL > 3.1..

c42f:
Sorry, I've been away on holiday and then back without much free time.

Thanks a lot for this patch, and nicely done working out the workarounds.  See #89 for the glsl version issues - it's a known problem, but I haven't had the time to fix it yet.

PR #163, opened 2017-01-24
Bugfix: V3d for bounding box offset to avoid draw quantization errors

c42f:
When using V3f, the bounding box position will be quantized to float32
precision far from the origin, resulting in visual errors where a
dataset doesn't fit within its supposed bounding box.


c42f:
@andyferris does this fix your problem?

andyferris:
Yes, perfect thank you :)

PR #165, opened 2017-03-29
DrawBBox for displaz

mrosen:
Here's a start at an application that allow a displaz user to interactively define a bounding box.  You can set the center point and grow the box by 50% in x, y and z directions by pressing the corresponding keys.  Others may find it useful as a starting point.

### DIFF
--- a/test/DrawBBox.py
+++ b/test/DrawBBox.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python2
+import subprocess 
+import numpy
+import sys
+
+# Edit this as required (windows, mac).  Be able to write this.
+PLY_FILE="/tmp/displaz_bbox.ply"
+
+# write .ply file representing a box with specfied center and transform to f
+def MakeBBox(v_center, transform, f):
+  f.write("ply\n")
+  f.write("format ascii 1.0\n")
+  f.write("comment simple rectangular prism\n")
+  f.write("element vertex 8\n")
+  f.write("property float x\n")
+  f.write("property float y\n")
+  f.write("property float z\n")
+  f.write("element face 2\n")
+  f.write("property list uint int vertex_index\n")
+  f.write("end_header\n")
+
+  # apply transform to unit vectors and then add result to center
+  unit_vectors = [ 
+    [0, 1, -1, 0], [1, 0, -1, 0], [0, -1, -1, 0], [-1, 0, -1, 0], 
+    [0, 1, 1, 0], [1, 0, 1, 0], [0, -1, 1, 0], [-1, 0, 1, 0]
+  ] 
+  xfm_unit_vectors = [transform.dot(unit) for unit in unit_vectors]
+  vertices = [numpy.add(v_center, xuv) for xuv in xfm_unit_vectors]
+
+  # write out resulting vertices
+  map(lambda v: f.write("%s\n" % " ".join(str(elem) for elem in v)), vertices)
+  f.write("4 0 1 2 3\n")
+  f.write("4 4 5 6 7\n")
+
+# write out the .ply file and have Displaz render it
+def UpdateDisplaz(v, T):
+    t = open(PLY_FILE, 'w')
+    t.truncate()
+    MakeBBox(v, T, t)
+    t.close()
+    p = subprocess.Popen(["displaz -script -rmtemp " + PLY_FILE], stdout=subprocess.PIPE, shell=True, bufsize=1)
+
+def Usage():
+  print "Usage:  DrawBBox <input point cloud file>\n"
+  print "  Shift Click will set the 3D cursor and put a 1 x 1 x 1 BBox there.\n"
+  print "   Do this first.\n"
+  print "  Then you can press x, y and z to grow it.\n"
+  print "  Remember that these keystrokes go into the Displayz application (needs focus)"
+  print "\n\nTry it:  press space.\n"
+
+def main():
+  Usage()
+  inputfile = ""
+  if (len(sys.argv) > 1):
+    inputfile = sys.argv[1]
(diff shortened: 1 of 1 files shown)


test/DrawBBox.py
```diff
@@ -0,0 +1,88 @@
...
+  # apply transform to unit vectors and then add result to center
+  unit_vectors = [ 
+    [0, 1, -1, 0], [1, 0, -1, 0], [0, -1, -1, 0], [-1, 0, -1, 0], 
+    [0, 1, 1, 0], [1, 0, 1, 0], [0, -1, 1, 0], [-1, 0, 1, 0]
+  ] 
+  xfm_unit_vectors = [transform.dot(unit) for unit in unit_vectors]
+  vertices = [numpy.add(v_center, xuv) for xuv in xfm_unit_vectors]
+
+  # write out resulting vertices
+  map(lambda v: f.write("%s\n" % " ".join(str(elem) for elem in v)), vertices)
```
c42f: As a small nicety, this might be somewhat clearer as a simple `for` loop (`map` would normally be used to gather output, but here it's being used more as a `foreach`, which is something python lacks):

```python
v_center = np.array(v_center)
for uvec in unit_vectors:
    vertex = v_center + transform.dot(uvec)
    f.write("%.17g %.17g %.17g\n" % tuple(vertex))
```

test/DrawBBox.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python2
+import subprocess 
+import numpy
+import sys
+
+# Edit this as required (windows, mac).  Be able to write this.
+PLY_FILE="/tmp/displaz_bbox.ply"
+
+# write .ply file representing a box with specfied center and transform to f
+def MakeBBox(v_center, transform, f):
```
c42f: Did you see the python bindings inside the `bindings/python` directory?  They're pretty basic, but some ply writing utilities like this one could arguably go in there rather than in this example.  Just a thought really, I don't mind having this here, for the moment.

test/DrawBBox.py
```diff
@@ -0,0 +1,88 @@
...
+  f.write("ply\n")
+  f.write("format ascii 1.0\n")
+  f.write("comment simple rectangular prism\n")
+  f.write("element vertex 8\n")
+  f.write("property float x\n")
+  f.write("property float y\n")
+  f.write("property float z\n")
+  f.write("element face 2\n")
+  f.write("property list uint int vertex_index\n")
+  f.write("end_header\n")
```
c42f: Minor note - writing literal blocks of text like this to file can be done neatly with the standard `textwrap` module (see `textwrap.dedent` to remove pesky indentation from a big block of text in a function).

test/DrawBBox.py
```diff
@@ -0,0 +1,88 @@
...
+  f.write("4 0 1 2 3\n")
+  f.write("4 4 5 6 7\n")
+
+# write out the .ply file and have Displaz render it
+def UpdateDisplaz(v, T):
+    t = open(PLY_FILE, 'w')
+    t.truncate()
+    MakeBBox(v, T, t)
+    t.close()
+    p = subprocess.Popen(["displaz -script -rmtemp " + PLY_FILE], stdout=subprocess.PIPE, shell=True, bufsize=1)
```
c42f: I suggest that you use the default `shell=False`, and pass the arguments as a list `["displaz", "-script", "-rmtemp", PLY_FILE]` when calling `Popen`.  This is more annoying to type, but is far more reliable whenever you have arguments which would need to be quoted when passed to the shell, for example, empty arguments, file names with whitespace, etc.

c42f: Also, how about labelling your dataset?  If you use the `-label Box` option, it will show up in the dataset view with a more readable name, rather than the temporary file name.

test/DrawBBox.py
```diff
@@ -0,0 +1,88 @@
...
+
+def Usage():
+  print "Usage:  DrawBBox <input point cloud file>\n"
+  print "  Shift Click will set the 3D cursor and put a 1 x 1 x 1 BBox there.\n"
+  print "   Do this first.\n"
+  print "  Then you can press x, y and z to grow it.\n"
+  print "  Remember that these keystrokes go into the Displayz application (needs focus)"
```
c42f: typo: Displayz

c42f: Thanks for the docs!

test/DrawBBox.py
```diff
@@ -0,0 +1,88 @@
...
+  xfm_unit_vectors = [transform.dot(unit) for unit in unit_vectors]
+  vertices = [numpy.add(v_center, xuv) for xuv in xfm_unit_vectors]
+
+  # write out resulting vertices
+  map(lambda v: f.write("%s\n" % " ".join(str(elem) for elem in v)), vertices)
+  f.write("4 0 1 2 3\n")
+  f.write("4 4 5 6 7\n")
+
+# write out the .ply file and have Displaz render it
+def UpdateDisplaz(v, T):
```
c42f: The indentation here is inconsistent with the other functions.  Personally I prefer the standard PEP-8 style, with four spaces per indentation level.

mrosen:
Wow. Thanks for the thorough review. I'm new to python but will address
these excellent suggestions over the next few days.

msr

c42f:
Great!

test/DrawBBox.py
c42f: By the way, I'm using `%.17g` here, because this will exactly preserve all the binary digits of any floating point number, while keeping the string short if possible.

mrosen:
Sorry for the long delay here.  There were internal company issues as well as other distractions.  I think the new code is PEP-8 compliant and certainly does a lot more.  I hope others find this useful!

PR #167, opened 2017-05-10
Fix appveyor build

evetion:
Removed extra platform folder in packaging
Set path to include Qt, so packaging can find Qt dlls

See my build https://ci.appveyor.com/project/evetion/displaz/build/1


c42f:
Nice, thanks for looking at this.  I rebased my old branch to clean up the diff here, but yes - if you can create a PR against master that would be preferable.

evetion:
New PR #168, closing this one.

PR #185, opened 2018-04-21
Add support for extended classification e.g LAS 1.4, etc.

pandu-rao:


src/las_io.cpp
```diff
@@ -264,9 +263,15 @@ bool PointArray::loadLas(QString fileName, size_t maxPointCount,
         *numReturns++ = point.number_of_returns_of_given_pulse;
 #       endif
         *pointSourceId++ = point.point_source_ID;
-        // Put flags back in classification byte to avoid memory bloat
-        *classification++ = point.classification | (point.synthetic_flag << 5) |
-                            (point.keypoint_flag << 6) | (point.withheld_flag << 7);
+
+        if (point.extended_classification) {
```
nigels-com: Seems like we ought to be checking for `point.extended_point_type` here, rather than merely non-zero extended classification?

pandu-rao: Good point. What do you recommend?

CCInc:
@c42f would we be able to get this merged? would be super helpful 😄

nigels-com:
Looks good to me.

PR #187, opened 2018-06-11
Formulate intensity tone mapping as Middle/Contrast

nigels-com:
For mapping intensity values to greyscale we currently specify an _exposure_ which is internally divided by 400.0.

For improved intuition, expose _middle_ intensity (default 400.0) which is the intensity value that will map to 50% output greyscale.

This _middle_ parameter is something we'd like to use elsewhere, indicating the expected range of intensity for some particular sensor.

### DIFF
--- a/shaders/las_points.glsl
+++ b/shaders/las_points.glsl
@@ -9,12 +9,13 @@ uniform mat4 modelViewProjectionMatrix;
 //------------------------------------------------------------------------------
 #if defined(VERTEX_SHADER)
 
-uniform float pointRadius = 0.1;   //# uiname=Point Radius; min=0.001; max=10
-uniform float trimRadius = 1000000;//# uiname=Trim Radius; min=1; max=1000000
-uniform float exposure = 1.0;      //# uiname=Exposure; min=0.001; max=10000
-uniform float contrast = 1.0;      //# uiname=Contrast; min=0.001; max=10000
-uniform int colorMode = 0;         //# uiname=Colour Mode; enum=Intensity|Colour|Return Index|Point Source|Las Classification|File Number
-uniform int selectionMode = 0;     //# uiname=Selection; enum=All|Classified|First Return|Last Return|First Of Several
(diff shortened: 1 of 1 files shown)


shaders/las_points.glsl
```diff
@@ -9,12 +9,13 @@ uniform mat4 modelViewProjectionMatrix;
...
 
-uniform float pointRadius = 0.1;   //# uiname=Point Radius; min=0.001; max=10
-uniform float trimRadius = 1000000;//# uiname=Trim Radius; min=1; max=1000000
-uniform float exposure = 1.0;      //# uiname=Exposure; min=0.01; max=10000
-uniform float contrast = 1.0;      //# uiname=Contrast; min=0.01; max=10000
-uniform int colorMode = 0;         //# uiname=Colour Mode; enum=Intensity|Colour|Return Index|Point Source|Las Classification|File Number
-uniform int selectionMode = 0;     //# uiname=Selection; enum=All|Classified|First Return|Last Return|First Of Several
+uniform float pointRadius = 0.1;    //# uiname=Point Radius; min=0.001; max=10
+uniform float trimRadius = 1000000; //# uiname=Trim Radius; min=1; max=1000000
+uniform float middle = 400.0;       //# uiname=Middle; min=0.01; max=100000
```
c42f, line 14:
```suggestion
uniform float middle = 400.0;       //# uiname=Reference Intensity; min=0.01; max=100000
```

c42f (commented):
Eek, I'm unsure why I never noticed this PR.

Agreed that the 400 is rather arbitrary and unlikely to be useful unless you're viewing Riegl lidar data.

Does `exposure` do anything after this change? I guess it should be removed.

shaders/las_points.glsl
nigels-com: Done, updated.

Screenshot from 2021-02-16 21-21-55

nigels-com:
> Does `exposure` do anything after this change?

Yes, it is still applicable to the other modes such as Colour.

PR #193, opened 2020-11-22
Qt5 Linux build fix

JDuchniewicz:
Fixes this error.
```
displaz/src/render/Annotation.cpp:22:30: warning: ‘int QFontMetrics::width(const QString&, int) const’ is deprecated: Use QFontMetrics::horizontalAdvance [-Wdeprecated-declarations]
   22 |     int width = fm.width(text);
```

### DIFF
--- a/src/render/Annotation.cpp
+++ b/src/render/Annotation.cpp
@@ -10,6 +10,7 @@
 #include <QColor>
 #include <QImage>
 #include <QPainter>
+#include <QPainterPath>
 
 /// Creates and returns a texture containing some text.
 static std::unique_ptr<Texture> makeTextureFromText(const QString& text)


src/render/Annotation.cpp
```diff
@@ -10,6 +10,7 @@
 #include <QColor>
 #include <QImage>
 #include <QPainter>
+#include <QPainterPath>
```
c42f, line 13: Why does adding this include fix this?

The deprecation warning indicates that use of `fm.width` should be changed?

JDuchniewicz: Sorry, copied the wrong error message.

```
/home/jduchniewicz/Packages/displaz/src/render/Annotation.cpp:36:18: error: aggregate ‘QPainterPath path’ has incomplete type and cannot be defined
   36 |     QPainterPath path;
      |                  ^~~~
```

c42f: Oh I see. In that case this change makes perfect sense, thanks!

PR #196, opened 2021-02-11
Adjust exposure and contrast minimums from 0.01 to 0.001 

nigels-com:
Intensity data at around 50k-60k needs less exposure than 0.01


c42f:
Sounds sensible to me, thanks.

PR #204, opened 2022-02-09
PLY vertex normals

nigels-com:
I have some triangle mesh PLY that includes per-vertex normals.

Aside from the GLSL shaders, this change is needed for reading the per-vertex normals
from a PLY file.


nigels-com:
@c42f Any concerns with this change?

PR #222, opened 2023-11-29
Fix PointSourceID Datatype

CCInc:
1. The datatype of pointsourceid is incorrectly read from the LAZ file as uint8 rather than uint16, resulting in truncated values
2. The shader's % operator did not work well with the new uint16 value, so the mod function had to be used
3. With a greater range of pointsourceid's potential values, the list of 7 colors was not enough to visualize distinctly, so more colors were added.
4. When building the project, the shaders were not being copied into the build directory, preventing the executable from working when run through the build dir, so the shaders are now copied each time the project is built.

### DIFF
--- a/shaders/las_points.glsl
+++ b/shaders/las_points.glsl
@@ -29,7 +29,7 @@ in vec3 color;
 in float distance;
 in int returnNumber;
 in int numberOfReturns;
-in int pointSourceId;
+in mediump int pointSourceId;
 in int classification;
 //in float heightAboveGround;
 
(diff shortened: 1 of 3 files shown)


shaders/las_points.glsl
```diff
@@ -29,7 +29,7 @@ in vec3 color;
 in float distance;
 in int returnNumber;
 in int numberOfReturns;
-in int pointSourceId;
+in mediump int pointSourceId;
```
nigels-com, line 32: Is this treating pointSourceId as a float, rather than an int?
Is _that_ why % modulo isn't working as expected?

CCInc: I think I was wrong about % not working. It seems to be fine now, I reverted those changes.

